PR #1, opened 2023-06-27
Handle an older version of Plotly which does not have Scatter.offsetg…

joshelser:
…roup

When not running locally, it appears that we get an older version of plotly which does not have the offsetgroup atribute on Scatter. However, the offsetgroup does not seem to be important since we want the bar and scatter plot to render on top of each other.

This change was introduced in plotly v5.12.0, so we must be running v5.11.0 or older.


joshelser:
@rymurr @jacques-n

rymurr (approved):

PR #2, opened 2023-06-27
Rely on project rules for maintainer approval to run workflow

joshelser:


joshelser:
Bah, and then i find docs that say this doesn't do what I want.

PR #7, opened 2023-06-30
Clean up devdeploy

joshelser:


.pre-commit-config.yaml
```diff
@@ -8,7 +8,9 @@ repos:
     hooks:
       - id: check-yaml
       - id: end-of-file-fixer
+        exclude: LICENSE.*
```
rymurr, line 11: I would rather fix than exclude

rymurr (approved):

PR #10, opened 2023-07-04
Remove devdeploy database argument and take it from config

anshuldata:


deploy/devdeploy.py
```diff
@@ -42,7 +42,7 @@ def _copy_opscenter_files(cur, schema: str, stage: str):
     cur.execute(body)
 
 
-def devdeploy(profile: str, database: str, schema: str, stage: str):
```
rymurr, line 45: should schema be the same pattern as database?

deploy/helpers.py
```diff
@@ -35,6 +35,8 @@ def remove_quotes(string):
     warehousename = remove_quotes(config.get(f"connections.{profile}", "warehousename"))
     dbname = remove_quotes(config.get(f"connections.{profile}", "dbname"))
 
+    if len(dbname) == 0:
+        raise ValueError("Database must be specified in the config string.")
```
rymurr, line 39: put the config string name instead. More descriptive

deploy/devdeploy.py
```diff
@@ -64,37 +63,31 @@ def devdeploy(profile: str, database: str, schema: str, stage: str):
 
 
 def usage():
-    print(
-        "devdeploy.py -p <snowsql_profile_name> [(-d | --database=) <database_name>] [(-s | --stage=) <stage_name>]"
-    )
+    print("devdeploy.py -p <snowsql_profile_name> [(-s | --stage=) <stage_name>]")
```
rymurr, line 66: I don't think we need to specify the stage name. Its as much set as the schemas. The only reason its an option in helpers is because in dev deploy the stage has to have a name whereas in prod its taken from the default stage (and the stage arg should be an empty string from prod deploy)

deploy/devdeploy.py
anshuldata: I didn't get this comment. Do you mean, just keep schema as some global variable with fix value and remove from function argument ?

deploy/helpers.py
anshuldata: done

deploy/devdeploy.py
anshuldata: done

anshuldata:
@rymurr 
I am not sure why test "PR Build Check / test-deploy" seems flaky (github actions shows failure in earlier occurences too).
I don't seem to find an option to re-trigger failed test. Any idea how to do it for this one ?

anshuldata:
> if you rebase the build should pass now

Even after rebase it is failing

rymurr:
@joshelser any idea whats up here?

rymurr:
hopefullt the break is fixed in #11

joshelser:
> hopefullt the break is fixed in https://github.com/sundeck-io/OpsCenter/pull/11

Yeah, that looks like it should do it. Nice find.

rymurr:
I've rebased for you @anshuldata . Will merge when it passes

PR #13, opened 2023-07-06
Increase default warehouse size of warehouse and query history mainte…

joshelser:
…nance serverless tasks

We've seen that an x-small warehouse can fail to complete once we're in the 10's of millions of rows for QUERY_HISTORY. Increase the default warehouse size and let snowflake scale it down as necessary.


bootstrap/090_post_setup.sql
```diff
@@ -34,14 +34,14 @@ BEGIN
 CREATE OR REPLACE TASK TASKS.WAREHOUSE_EVENTS_MAINTENANCE
     SCHEDULE = '60 minute'
     ALLOW_OVERLAPPING_EXECUTION = FALSE
-    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "XSMALL"
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "LARGE"
```
rymurr, line 37: I think this one can stay xsmall. It appears to complete very quickly even on xsmall

rymurr (approved):

joshelser:
Doron is chasing the new Cypress tests failure.

PR #14, opened 2023-07-06
Improve cypress test flakyness

doronrosenberg:
Add more timeouts to the test to ensure they wait for steamlit to start up properly.

### DIFF
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -30,11 +30,19 @@ jobs:
         run: |
           mkdir $HOME/.snowsql
           envsubst < deploy/opscenter.config > $HOME/.snowsql/config
-      - name: set opscenter env
+      - name: set env
         run: |
-          echo "OPSCENTER_PACKAGE=PACKAGE_DEV" >> "$GITHUB_ENV"
-          echo "OPSCENTER_DATABASE=\"DATABASE_OPSCENTER_DEV\"" >> "$GITHUB_ENV"
-          echo "OPSCENTER_APP=APP_OPSCENTER_DEV" >> "$GITHUB_ENV"
+          suffix=$(echo ${GITHUB_HEAD_REF} | sed "s/\./_/g" | sed "s/-/_/g" | sed "s\/\_\g")_${GITHUB_RUN_ID}
+          echo "OPSCENTER_PACKAGE=PACKAGE_${suffix}" >> "$GITHUB_ENV"
+          echo "OPSCENTER_DATABASE=\"DATABASE_${suffix}\"" >> "$GITHUB_ENV"
+          echo "OPSCENTER_APP=APP_${suffix}" >> "$GITHUB_ENV"
+      - name: run deploy
+        env:
+          OPSCENTER_PACKAGE: ${{env.OPSCENTER_PACKAGE}}
+          OPSCENTER_DATABASE: ${{env.OPSCENTER_DATABASE}}
+          OPSCENTER_APP: ${{env.OPSCENTER_APP}}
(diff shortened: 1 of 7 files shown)


rymurr (approved):

.github/workflows/cypress.yaml
```diff
@@ -22,20 +22,28 @@ jobs:
...
           envsubst < deploy/opscenter.config > $HOME/.snowsql/config
-      - name: set opscenter env
+      - name: set env
         run: |
-          echo "OPSCENTER_PACKAGE=PACKAGE_DEV" >> "$GITHUB_ENV"
-          echo "OPSCENTER_DATABASE=\"DATABASE_OPSCENTER_DEV\"" >> "$GITHUB_ENV"
-          echo "OPSCENTER_APP=APP_OPSCENTER_DEV" >> "$GITHUB_ENV"
-      - name: run opscenter streamlit
+          suffix=$(echo ${GITHUB_HEAD_REF} | sed "s/\./_/g" | sed "s/-/_/g" | sed "s\/\_\g")_${GITHUB_RUN_ID}
+          echo "OPSCENTER_PACKAGE=PACKAGE_${suffix}" >> "$GITHUB_ENV"
```
rymurr, line 36: we prob want to chang ethis to 
`echo "OPSCENTER_PACKAGE=PACKAGE_CYPRESS_${suffix}" >> "$GITHUB_ENV"`
similar to the ones below
otherwise this job may step on the other CI job (which uses the same suffix for its deploy)

.github/workflows/cypress.yaml
```diff
@@ -22,20 +22,28 @@ jobs:
...
+          echo "OPSCENTER_DATABASE=\"DATABASE_${suffix}\"" >> "$GITHUB_ENV"
+          echo "OPSCENTER_APP=APP_${suffix}" >> "$GITHUB_ENV"
+      - name: run deploy
+        env:
+          OPSCENTER_PACKAGE: ${{env.OPSCENTER_PACKAGE}}
```
rymurr, line 41: I don't think any of these env variables are used in devdeploy... @joshelser what do you think?

.github/workflows/cypress.yaml
doronrosenberg: Is `PACKAGE_CYPRESS_${suffix}` a value that exists?

rymurr: thats the value. the key is OPSCENTER_PACKAGE. Or am I misunderstanding something

rymurr (approved):
lets try i tout!

rymurr (approved):

.github/workflows/cypress.yaml
joshelser: Correct, they're not used in devdeploy which is what lead to some confusion.

This is the next level of cleanup i'd do between deploy.py and devdeploy.py (and intentionally avoided in case we needed to do any urgent 'hotfixes' of opscenter). We should be able to combine some of these functions now that we have these two entrypoints closer together.

joshelser (approved):
The pre-commit checks are going to fail because we have changed the workflow but the pull request is going to pull the workflow from the base-branch `main`. The changes look good to me. I'm in favor of merging and we just tweak as necessary.

.github/workflows/pull-request.yaml
```diff
@@ -3,7 +3,9 @@ name: PR Build Check
 on:
   pull_request_target:
     types: [opened, synchronize, reopened]
-
+  push:
+    branches:
+      - '*'
```
joshelser, line 8: revert this too, please.

joshelser (changes requested):

joshelser (approved):

PR #15, opened 2023-07-07
only compile condition, dont run it on all data

rymurr:
When creating a label we were running validate on the entire label history. This will only ensure the label will compile and will drastically improve perf of create label commands.


joshelser (approved):
LGTM

PR #16, opened 2023-07-07
"Probes" page sanity tests

vqmarkman:
- "Probes" page sanity tests.
- Generic way to press the button
- Generic way to fill in the page
- Function to create unique sting name (for probes and labels)

### DIFF
--- a/cypress/e2e/probes.spec.cy.js
+++ b/cypress/e2e/probes.spec.cy.js
@@ -1,4 +1,4 @@
-import { checkNoErrorOnThePage, setup, fillInProbeForm, buttonClick, buttonCheckExists, generateUniqueName } from "../support/utils";
+import { checkNoErrorOnThePage, setup, fillInProbeForm, buttonClick, buttonCheckExists, generateUniqueName, probeDelete } from "../support/utils";
 
 describe("Probes section", () => {
   before(() => {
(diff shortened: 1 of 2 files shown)


joshelser:
You'll need to rebase your PR to get the existing Cypress Tests job to pass.

Pull requests will still run a workflow based on what is in `main` currently, so this PR will have to be merged before we see the newly added tests running.

doronrosenberg (approved):

vqmarkman:
@joshelser , my PR was rebased, tests ran, but one thing I don’t understand is: you said that only tests checked into main will run. This does not appear to be the case: **probes.spec.cy.js** only exists in my branch in my fork.  

Screenshot 2023-07-07 at 4 53 44 PM

Failures:  
(1) I've never seen failure in probes like in this run.
(2) Is a known issue. I'm going to separate warehouses.spec.js into two separate tests and will skip "Heatmap" that appears to cause the problem.

cc @doronrosenberg

joshelser:
> one thing I don’t understand is: you said that only tests checked into main will run. This does not appear to be the case: probes.spec.cy.js only exists in my branch in my fork.

Ah, sorry. I was not specific enough. The Github workflow files are "selected" from `main`. Since the cypress workflow automatically finds all of the .spec.cy.js files to run automatically, those are automatically loaded in the PR (without having to wait). I was a little confused after the back and forth fixing the workflow last week :)

vqmarkman:
Discarding this PR, new one is here: https://github.com/sundeck-io/OpsCenter/pull/22

vqmarkman:
Discarding this PR, new one is here: https://github.com/sundeck-io/OpsCenter/pull/22

PR #17, opened 2023-07-07
Pass through snowsql config to devdeploy for cypress workflow

joshelser:
empty PR to make sure the workflow changes from #14  are working as we expect


joshelser:
> empty PR to make sure the workflow changes from https://github.com/sundeck-io/OpsCenter/pull/14 are working as we expect

It did not work. `devdeploy` defaults to a `local_opscenter` but the snowsql config we build was `opscenter`

doronrosenberg (approved):

PR #21, opened 2023-07-07
Fix cypress setup ordering

doronrosenberg:


joshelser:
this, coupled with a permission change on the snowflake user, should fix the cypress setup 🤞🏼

PR #24, opened 2023-07-11
Settings page test

vqmarkman:


cypress/e2e/settings.spec.cy.js
```diff
@@ -0,0 +1,67 @@
...
+
+    cy.wait(5000);
+
+    cy.get('button[role="tab"]', {timeout: 20000})
+      .should("exist")
+      .contains("Config")
+      .should("exist")
+      .click()
+
+    fillInTheSettingsConfigForm(10.00, 20.00, 30.00);
```
joshelser, line 35: Should there be a check to verify that these properties are now reflected on the page after save?

vqmarkman: yes, there should be. 
These are sanity tests that validate there are no exceptions or errors on the pages during object creation/deletion.
Next set of tests are (a) Correctness checks. (b) Error conditions

joshelser: > These are sanity tests that validate there are no exceptions or errors on the pages during object creation/deletion.

Gotcha. Sorry, the intended scope of the test was not apparent to me. Can you make sure there is an issue opened for that work (so we know when it's coming)? Thanks!

PR #46, opened 2023-07-12
Better wait logic for app materialization

### LINKED ISSUE #45: Better wait logic for app materialization
Right now we wait a minute to let the app materialization complete in Cypress, move that to the gitub action to only wait once per test run.

doronrosenberg:
Move wait logic to cypress action.

Closes #45

### DIFF
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -41,6 +41,8 @@ jobs:
         run: "poetry run python deploy/cypress_setup.py -p opscenter"
       - name: run streamlit
         run: "poetry run streamlit run app/ui/Home.py &"
+      - name: Wait one minute for materialization to happen
+        run: sleep 60
       - name: Cypress run
         uses: cypress-io/github-action@v4
         with:
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -41,8 +41,6 @@ jobs:
         run: "poetry run python deploy/cypress_setup.py -p opscenter"
       - name: run streamlit
         run: "poetry run streamlit run app/ui/Home.py &"
-      - name: Wait one minute for materialization to happen
-        run: sleep 60
       - name: Cypress run
         uses: cypress-io/github-action@v4
         with:
--- a/deploy/cypress_setup.py
+++ b/deploy/cypress_setup.py
@@ -27,13 +27,14 @@ def cypress_setup(profile: str, schema: str):
         try:
             # Execute a query to fetch data from the table
             cur.execute(
-                "SELECT * FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE');"
+                "SELECT count(*) FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE') and value is not null ;"
             )
 
             rows = cur.fetchall()
 
             # if we have two rows, means materialization is complete
             if len(rows) == 2:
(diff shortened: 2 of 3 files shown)


.github/workflows/cypress.yaml
```diff
@@ -41,6 +41,8 @@ jobs:
         run: "poetry run python deploy/cypress_setup.py -p opscenter"
       - name: run streamlit
         run: "poetry run streamlit run app/ui/Home.py &"
+      - name: Wait one minute for materialization to happen
+        run: sleep 60
```
joshelser, line 45: In the same way that we codified some setup in `cypress_setup.py`, is it possible we check for a discrete condition via SQL objects rather than a blind wait?

It looks like the `INTERNAL.TASK_QUERY_HISTORY` and `INTERNAL.TASK_WAREHOUSE_EVENTS` tables might already contain the information we'd need.

doronrosenberg: I opened #47 to see if that is possible, but we could try to do that now.  How would we know exactly that the data is there, look at the count?

deploy/cypress_setup.py
```diff
@@ -20,7 +21,32 @@ def cypress(profile: str, schema: str):
...
     )
 
-    conn.close()
+    start_time = time.time()
+
+    while True:
+        try:
+            # Execute a query to fetch data from the table
+            cur.execute(
+                "SELECT * FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE');"
```
vqmarkman, line 30: Can't comment on correctness of the check (don't know if say value could be null and it means we are not setup correctly yet). Maybe we should protect against that? Or empty string?

`SELECT count(*) FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE') and value is not null ;`

deploy/cypress_setup.py
```diff
@@ -20,7 +21,32 @@ def cypress(profile: str, schema: str):
...
+            )
+
+            rows = cur.fetchall()
+
+            # if we have two rows, means materialization is complete
+            if len(rows) == 2:
+                break
```
vqmarkman, line 37: It would be nice to print a message that indicates that we've detected successful materialization.

deploy/cypress_setup.py
```diff
@@ -20,7 +21,32 @@ def cypress(profile: str, schema: str):
...
+
+        finally:
+            cur.close()
+            conn.close()
+
+        elapsed_time = time.time() - start_time
+        # bail after 3 minutes
+        if elapsed_time >= 150:
+            break
```
vqmarkman, line 46: Same here: we should say that setup has not finished in allocated time.

vqmarkman (changes requested):
@doronrosenberg, thanks for fixing it! 

I think this script needs to exit with non-zero error code in the case when setup was not complete in time, so we don't continue to the next step in the GH workflow and fail right away without even attempting to run tests. What do you think?

deploy/cypress_setup.py
```diff
@@ -20,7 +21,34 @@ def cypress(profile: str, schema: str):
...
+        try:
+            # Execute a query to fetch data from the table
+            cur.execute(
+                "SELECT count(*) FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE') and value is not null ;"
+            )
+
+            rows = cur.fetchall()
+
+            # if we have two rows, means materialization is complete
+            if len(rows) == 2:
```
vqmarkman, line 36: Output from count(*) is a single value. I don't think this check is correct.

vqmarkman (approved):
ship it!

PR #49, opened 2023-07-12
Add cypress tests for initial setup

### LINKED ISSUE #48: Add tests for initial setup
Add cypress tests for initial setup

doronrosenberg:
Covers:
  - ensure the first step is marked as completed
  - ensure the Sundeck link is correct

Closes #48.


doronrosenberg:
Cypress will fail because the env variable is being set in the cypress.yaml but we are using the main cypress.yaml so its missing and the test fails

joshelser (approved):

PR #51, opened 2023-07-13
#70 Enable notifications using snowflake SSO integration

scgkiran:
After installation, if user decides to enable notifications, today one can do it by Signing up for a free Sundeck account using email-Id. These changes are to provide another way to do the same using Snowflake SSO. In OpsCenter UI an SQL script will be provided which needs to be run as account admin to enable notifications using Snowflake SSO integration.

### DIFF
--- a/app/ui/Home.py
+++ b/app/ui/Home.py
@@ -32,3 +32,10 @@ else:
     An overview dashboard will arrive here soon!
     """
     )
+    if config.has_tenant_url():
+        tenant_url = config.get_tenant_url()
+        st.markdown(
+            f"""
+                [Go to my Sundeck account]({tenant_url})
+            """
+        )
--- a/app/ui/Home.py
+++ b/app/ui/Home.py
@@ -36,6 +36,8 @@ else:
         tenant_url = config.get_tenant_url()
         st.markdown(
             f"""
+                To explore the Sundeck account, right-click on the link below and choose "Open link in new tab/window."
+
                 [Go to my Sundeck account]({tenant_url})
             """
         )
--- a/app/ui/setup.py
+++ b/app/ui/setup.py
@@ -112,7 +112,7 @@ END;
         )
 
         option1, option2 = st.tabs(
-            ["Option 1: Signup with Snowflake SSO", "Option 2: Signup using email-Id"]
+            ["Option 1: Signup with Snowflake SSO", "Option 2: Signup with email"]
         )
         with option1:
             sundeck_signup_with_snowflake_sso(
@@ -201,7 +201,7 @@ def sundeck_signup_with_snowflake_sso(
     st.markdown(
         """
         #### Sundeck-signup using Snowflake-SSO. Follow these steps to enable notifications:
-        1. Prepare to create a Sundeck account. This is needed to for step 2.
+        1. Prepare to create a Sundeck account. This is needed for step 2.
 
         """
     )
@@ -209,7 +209,7 @@ def sundeck_signup_with_snowflake_sso(
     api_integration_name = "OPSCENTER_SUNDECK_EXTERNAL_FUNCTIONS_SSO"
     if st.button("Enable Sundeck API Integration", key="create_api_integration"):
         req = perms.request_aws_api_integration(
-            "opscenter_api_integration_sso",
+            "opscenter_sso_api_integration",
             (ef_url,),
             perms.AwsGateway.API_GATEWAY,
             OPSCENTER_ROLE_ARN,
@@ -218,14 +218,13 @@ def sundeck_signup_with_snowflake_sso(
             None,
         )
         if req is None:
-            print("Successfully created API integration.")
             st.code(
                 """
                 "Successfully created API integration. Please run the following command in your Snowflake account"
                 """
             )
         else:
-            print("Please run the following command in your Snowflake account:")
+            st.text("Please run the following command in your Snowflake account:")
             setup_func = "ADMIN.SETUP_REGISTER_TENANT_FUNC()"
             gateway_prefixes = f"'{ef_url}'"
             st.code(
@@ -263,11 +262,14 @@ def generate_security_integration_code(
     sf_region: str, sd_deployment: str, name: str
 ) -> str:
     redirect_url = get_redirect_url_for_security_integration(sf_region, sd_deployment)
-    return (
-        f"create security integration if not exists {name} type=oauth enabled=true oauth_client=CUSTOM "
-        + f"oauth_client_type='CONFIDENTIAL' oauth_redirect_uri='{redirect_url}' oauth_issue_refresh_tokens=true "
-        + "oauth_refresh_token_validity=86400 pre_authorized_roles_list = ('PUBLIC'); "
-    )
+    return f"create security integration if not exists {name} " \
+        + "type=oauth " \
+        + "enabled=true oauth_client=CUSTOM " \
+        + "oauth_client_type='CONFIDENTIAL' " \
+        + f"oauth_redirect_uri='{redirect_url}' " \
+        + "oauth_issue_refresh_tokens=true " \
+        + "oauth_refresh_token_validity=86400 " \
+        + "pre_authorized_roles_list = ('PUBLIC'); "
 
 
 def generate_register_tenant_code(db: str, security_integration_name: str) -> str:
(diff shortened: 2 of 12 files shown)


joshelser:
@scgkiran thanks for the (draft) PR. Can you please make a Github issue which captures what you're planning to do so others can follow along and give feedback if they choose? Thanks!

rymurr (commented):
What is the behaviour for someone who updates the app from an older version and hasn't (or has) logged into the notification center. I assume it still works? Can they switch to SSO?

app/ui/Home.py
```diff
@@ -32,3 +32,10 @@
     An overview dashboard will arrive here soon!
     """
     )
+    if config.has_tenant_url():
+        tenant_url = config.get_tenant_url()
+        st.markdown(
+            f"""
+                [Go to my Sundeck account]({tenant_url})
```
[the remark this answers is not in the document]
scgkiran: done

app/ui/setup.py
```diff
@@ -97,60 +87,84 @@ def expander(num: int, title: str, finished: bool) -> st.expander:
...
+    token_input = st.text_input("Sundeck Token", key="token")
+    if st.button("Enable Notifications", key="connect"):
+
+        msg = st.empty()
+        msg.warning("Connecting. Please do not navigate away from this page.")
+        token, url = decode_token(token_input)
+        connection.Connection.get().call("INTERNAL.SETUP_SUNDECK_TOKEN", url, token)
+        req = perms.request_aws_api_integration(
+            "opscenter_api_integration",
+            (
```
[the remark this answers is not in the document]
scgkiran: done

app/ui/setup.py
```diff
@@ -172,3 +186,205 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+        if sundeck_region == "us-east-1":
+            return stage_url_map[sundeck_region]
+        else:
+            return stage_url_map["us-west-2"]
+    else:
+        raise Exception("Invalid deployment")
+
+
+def get_sundeck_region(sf_region: str) -> str:
+    # Supported Sundeck Regions ["us-east-1", "us-east-2.aws", "us-west-2"]
```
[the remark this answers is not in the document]
scgkiran: Added some detail, please check

deploy/deploy.py
```diff
@@ -150,13 +151,18 @@ def _install_or_update_package(
...
-    opts, args = getopt.getopt(argv, "xhsp:v:", ["profile=", "version="])
+    opts, args = getopt.getopt(
+        argv, "xhsd:p:v:", ["deployment=", "profile=", "version="]
+    )
     for opt, arg in opts:
         if opt == "-h":
             print(
-                "deploy.py -p <snowsql_profile_name> -s (skip install) -x (skip package creation/upload entirely) -v <version_name> --profile <snowsql_profile_name>, --version <version_name>"
+                "deploy.py -p <snowsql_profile_name> -s (skip install) -x (skip package creation/upload entirely) "
+                + "-v <version_name>  -d <sundeck_deployment> --profile <snowsql_profile_name>, "
```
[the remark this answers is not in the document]
scgkiran: done

scgkiran:
> What is the behaviour for someone who updates the app from an older version and hasn't (or has) logged into the notification center. I assume it still works? Can they switch to SSO?

It should work. The changes to switch to SSO are not ready. Switch will be through a sundeck ddl.

app/ui/Home.py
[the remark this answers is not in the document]
joshelser: I see you updated down in setup.py. Did you miss updating this on too, Chandra?

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+            f"""
+                #### Sundeck account is created
+                To visit Sundeck UI, [right click here]({tenant_url}) and open this link in a new tab/window.
+            """
+        )
+
+    st.markdown(
+        """
+        #### Sundeck-signup using Snowflake-SSO. Follow these steps to enable notifications:
+        1. Prepare to create a Sundeck account. This is needed to for step 2.
```
joshelser, line 204:
```suggestion
        1. Prepare to create a Sundeck account. This is needed for step 2.
```

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+        )
+        if req is None:
+            print("Successfully created API integration.")
+            st.code(
+                """
+                "Successfully created API integration. Please run the following command in your Snowflake account"
+                """
+            )
+        else:
+            print("Please run the following command in your Snowflake account:")
```
joshelser, line 228: Should this be an `st.text` call (or similar) or simply removed?

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+            "opscenter_api_integration_sso",
+            (ef_url,),
+            perms.AwsGateway.API_GATEWAY,
+            OPSCENTER_ROLE_ARN,
+            None,
+            api_integration_name,
+            None,
+        )
+        if req is None:
+            print("Successfully created API integration.")
```
joshelser, line 221: `st.text` instead of `print` or just remove?

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+
+
+def generate_security_integration_code(
+    sf_region: str, sd_deployment: str, name: str
+) -> str:
+    redirect_url = get_redirect_url_for_security_integration(sf_region, sd_deployment)
+    return (
+        f"create security integration if not exists {name} type=oauth enabled=true oauth_client=CUSTOM "
+        + f"oauth_client_type='CONFIDENTIAL' oauth_redirect_uri='{redirect_url}' oauth_issue_refresh_tokens=true "
+        + "oauth_refresh_token_validity=86400 pre_authorized_roles_list = ('PUBLIC'); "
```
joshelser, line 269: nit: a big f-string multi-line string here, too? I would also put each attribute to `create security integration` on its own line for readability
```
f"""
...
"""
```

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+    return (
+        f"""
+BEGIN
+    CREATE OR REPLACE API INTEGRATION {api_integration_name} api_provider = aws_api_gateway """
+        + f""" api_aws_role_arn = '{OPSCENTER_ROLE_ARN}' api_allowed_prefixes = ({gateway_prefixes}) enabled = true;
+    GRANT USAGE ON INTEGRATION {api_integration_name} TO APPLICATION "{app_name}";
+    CALL {setup_func};
+END;
+"""
+    )
```
joshelser, line 294: nit unnecessary parens?
```suggestion
    return
        f"""
BEGIN
    CREATE OR REPLACE API INTEGRATION {api_integration_name} api_provider = aws_api_gateway """
        + f""" api_aws_role_arn = '{OPSCENTER_ROLE_ARN}' api_allowed_prefixes = ({gateway_prefixes}) enabled = true;
    GRANT USAGE ON INTEGRATION {api_integration_name} TO APPLICATION "{app_name}";
    CALL {setup_func};
END;
"""
```

app/ui/setup.py
```diff
@@ -172,3 +184,208 @@ def sndk_url(account: str, user: str, region: str) -> str:
...
+        "AWS_EU_WEST_2": "eu-west-2",
+        "AWS_AP_NORTHEAST_2": "ap-northeast-2",
+        "AWS_EU_NORTH_1": "eu-north-1",
+        "AWS_AP_NORTHEAST_3": "ap-northeast-3",
+        "AWS_SA_EAST_1": "sa-east-1",
+        "AWS_EU_WEST_3": "eu-west-3",
+        "AWS_AP_SOUTHEAST_3": "ap-southeast-3",
+        "AWS_US_GOV_WEST_1": "us-gov-west-2",
+        "AWS_US_EAST_1_GOV": "us-east-1",
+    }
```
joshelser, line 389: I find myself wishing these two maps were built from the same key-set. When Snowflake inevitably adds a new region, we would want to make sure both of these `dicts` are updated.

Could we create these dicts once and just expose the functions? I would love a comment to remind developers later to make sure both maps are updated (unless you can come up with a good way to make it hard for a developer to accidentally add a region to one of the dicts and not the other :) )

bootstrap/021_external_functions.sql
```diff
@@ -104,19 +104,71 @@ create function if not exists internal.get_ef_url()
...
+    returns object
+    immutable
+as
+$$
+    iff(length(internal.ef_register_tenant(request):error) != 0,
+        internal.throw_exception(internal.ef_register_tenant(request):error),
+        internal.ef_register_tenant(request))::object
+$$;
+
+create or replace function tools.register_tenant(client_id varchar, client_secret varchar)
```
joshelser, line 148: I believe that the `TOOLS` schema is exposed to users with functions that users can expect to use as OpsCenter users (e.g. tools that help the users understand and user snowflake better). This seems more like an implementation detail of setting up the Sundeck account. Could we move this into `INTERNAL`?

IMO we should strive for a no-arg procedure in `ADMIN` that would create the necessary objects and then call the external function. If the account was already set up, it can return whatever is necessary; otherwise, it's created for the user.

app/ui/setup.py
scgkiran: precommit (black) is not happy about it. It is adding parens

app/ui/setup.py
joshelser: Ok, i was only thinking about readability for devs, not presentation to users.

app/ui/Home.py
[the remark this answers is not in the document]
joshelser: Hrm. I don't know exactly what to check to determine if the sandbox is being applied to new windows. My worry would be that if the user does not explicitly force a new window/tab to be opened, the existing sandbox would be applied to it.

@doronrosenberg is there an easy check to verify if the sandbox applies to a new window/tab via devtools or something? (short of trying to submit a form and it failing)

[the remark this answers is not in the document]
rymurr: We definitely should not try to detect we are in a sandbox. Please add the right-click here too @scgkiran

joshelser (approved):
LGTM.

If the sandbox does apply when streamlit is opening a new window, we can add the hint to explicitly open link in a new tab/window in a follow-on.

app/ui/Home.py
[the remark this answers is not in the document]
joshelser: > We definitely should not try to detect we are in a sandbox.

to be clear, I meant as the user of opscenter, not inside of opscenter itself -- if streamlit is opening a new tab and we know that the sandbox isn't being applied, we can avoid the hint here. I don't recall if we were getting a new tab/window last time we hit this issue.

anshuldata (approved):
lgtm

Issue #38, opened 2023-07-12
Improve long/fixed timeouts in Cypress tests


joshelser:
We have a number of long timeouts to minimize flakiness in the Cypress tests because a UI element is not updated quickly enough. The downside is that this is onerous to write such tests, but also drives up the cost of pre-commit testing (because the tests have to run for a significantly longer period of time).

vqmarkman:
These commits should address most of the issues. Every failing test after that should be debugged separately.
I will keep an eye on the runs. Closing.

Issue #35, opened 2023-07-12
Exception "index 0 is out of bounds for axis 0 with size 0" seen on Task Settings Page


joshelser:
```
2023-07-06 21:22:17.107 Uncaught app exception
Traceback (most recent call last):
  File "$HOME/Library/Caches/pypoetry/virtualenvs/opscenter-g7Dx2M1v-py3.8/lib/python3.8/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 552, in _run_script
    exec(code, module.__dict__)
  File "$HOME/OpsCenter/app/ui/pages/10_Settings.py", line 109, in <module>
    setup.setup_block()
  File "$HOME/OpsCenter/app/ui/setup.py", line 54, in setup_block
    connection.execute(
IndexError: index 0 is out of bounds for axis 0 with size 0
```

It doesn't appear that this should be able to happen under normal circumstances. The hunch is that it is related to interacting with the page before Streamlit has completely loaded the page.

vqmarkman:
Another similar issue that we will track here is: "Exception when clicking on Warehouses tab/Heatmap and choosing 7 day range"

```
2023-07-06 16:29:37.456 Uncaught app exception
Traceback (most recent call last):
  File "/Users/vickymarkman/Library/Caches/pypoetry/virtualenvs/opscenter-g7Dx2M1v-py3.8/lib/python3.8/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 552, in _run_script
    exec(code, module.__dict__)
  File "/Users/vickymarkman/git-repo/OpsCenter/app/ui/pages/01_Warehouses.py", line 13, in <module>
    reports.display(options)
  File "/Users/vickymarkman/git-repo/OpsCenter/app/ui/reports.py", line 30, in display
    options[report](filter_values, credit_cost)
  File "/Users/vickymarkman/git-repo/OpsCenter/app/ui/reports_heatmap.py", line 31, in heatmap
    df["UTILIZATION"] = (df["UTILIZATION"].astype(float) * 100).round(1)
  File "/Users/vickymarkman/Library/Caches/pypoetry/virtualenvs/opscenter-g7Dx2M1v-py3.8/lib/python3.8/site-packages/pandas/core/frame.py", line 3761, in __getitem__
    indexer = self.columns.get_loc(key)
[...]

PR #52, opened 2023-07-13
Issue #38: (part 2) set the defaultCommandTimeout to 20 seconds.

vqmarkman:
(Time, in milliseconds, to wait until most DOM based commands are considered timed out.)


vqmarkman:
We have last 3 hardcoded waits on the "Settings" tabs. My attempt to remove them was not successful.
I keep running into exception (Issue #35)  
After this improvement, every single failure in these e2e tests needs to be looked at, debugged and fixed.

joshelser (approved):

PR #54, opened 2023-07-14
Basic, anonymized telemetry for page views and actions taken

joshelser:
One provided implementation which sends the events to an external function, but would be straightforward to implement any other.

### DIFF
--- a/app/ui/telemetry.py
+++ b/app/ui/telemetry.py
@@ -30,7 +30,7 @@ def send_telemetry(event: str, data: str):
             "select internal.telemetry(%(event)s, %(data)s)",
             {"event": event, "data": data},
         )
-    except Exception as e:
+    except Exception:
         success = False
         pass
     print(f"Telemetry: {event} {data}, {'success' if success else 'failed'}")


bootstrap/021_external_functions.sql
```diff
@@ -156,3 +201,8 @@ BEGIN
       INSERT (key, value)
       VALUES (source.key, source.value);
 END;
+
+CREATE OR REPLACE PROCEDURE admin.setup_external_functions() RETURNS STRING LANGUAGE SQL AS
+BEGIN
+    call admin.setup_external_functions('reference(\'opscenter_api_integration\')');
+END;
```
joshelser, line 208: I'm not convinced this will do what I want it to do, but I'm trying to fix the `devdeploy.py` setup where you have to create your own API integration via the scripting block.

These changes work for me to set my own api integration name and then pass that into the `setup_external_functions` procedure, but I fear I broke the `reference('opscenter_api_integration')` path through deploy.py.

joshelser:
Overtaken by #170

Issue #77, opened 2023-07-25
Rename cypress_setup.py to finish_local_setup.py and document necessity to call for local dev


JamesRTaylor:
There's a `cypress_setup.py` script that performs the required grants on the database for local development. As far as I can tell, it doesn't have anything to do with cypress, but it is necessary to call it for local development of OpsCenter. I propose we rename it to `finish_local_setup.py` and update the readme to indicate when it's necessary to call it.

JamesRTaylor:
Based on review feedback, we ended up combining cypress_setup.py with devdeploy.py so that a single script does everything

PR #78, opened 2023-07-25
Issue #77 Rename cypress_setup.py and document need to call for local dev

JamesRTaylor:


README.md
```diff
@@ -73,6 +73,9 @@ Then run the following commands:
 # Create dev database
 poetry run python deploy/devdeploy.py -p local_dev
 
+# Finish local setup
```
joshelser, line 76:
```suggestion
# Finish local setup sets internal state to mimic a set-up app (e.g. materializes data, starts tasks)
```

I'm sensitive to this "magical" script which hides that there is additional setup required (e.g. the fact that materialized data isn't actually present when a user installs the app). I'm OK with it to make it easier for developer productivity, but just want to make sure we're clear what the script is doing.

joshelser (changes requested):
You'll also have to change the github workflow which references `cypress_setup.py` Ironically, making this change will result in your PR failing precommit, but that is because Github always uses the workflow definition from `main` and not the pull-request branch (as a security mechanism).

JamesRTaylor:
Thanks for the review, @joshelser. I've pushed those updates

rymurr:
Imho if this really should be ran all the time it should be part of the devdeploy script. It feels odd to have two separate scripts for getting to where a developer can be productive

JamesRTaylor:
Agreed, @rymurr. I'll make that change too.

joshelser (approved):
LGTM. Makes me wonder if we'll next ask for a "skip local setup" flag on devdeploy, but that's an easy thing to add if the need arises 😆 

Thanks James!

joshelser:
cypress-test failure is as expected. Merging.

deploy/devdeploy.py
```diff
@@ -60,6 +100,10 @@ def devdeploy(profile: str, schema: str, stage: str, deployment: str):
     # Deploy the OpsCenter code into the stage.
     _copy_opscenter_files(cur, conn.schema, stage, deployment)
 
+    # Finish local setup by setting internal state to mimic a set-up app
+    # (e.g. materializes data, starts tasks)
+    _finish_local_setup(cur, conn.database, conn.schema)
```
rymurr, line 105: @JamesRTaylor can you put this behind a command line flag? There are instances where we don' twant this to happen...eg when testing the refresh tasks

JamesRTaylor: Will do in a follow up PR

Issue #76, opened 2023-07-25
Labels negative test failed with the failure to find an element (this is a test problem and not a product issue)


vqmarkman:
Failed here: https://github.com/sundeck-io/OpsCenter/actions/runs/5659211804 
See attached video

https://github.com/sundeck-io/OpsCenter/assets/104043363/f2ab55b5-e62a-4647-9a87-c0830dfe364d

The problem in this test is that when "Cancel" button is clicked and we are brought back to the list of labels, cypress can't find label element on the page. I thought I resolved this issue by reloading the page, but it does not seem to be the case.

Need to find different a solution.

joshelser:
Thanks for filing this, Vicky! If you aren't able to get to the bottom of this today, could you please temporarily revert #58 or skip these tests that are now failing?

I also noticed a settings.spec.cy.js failure which looks to be similar to these label failures in https://github.com/sundeck-io/OpsCenter/actions/runs/5667117995/job/15355204886

vqmarkman:
@joshelser , I just submitted PR: https://github.com/sundeck-io/OpsCenter/pull/81 to take flaky tests out until I figure out what is causing that.
"Settings" failure is something different that I'm still looking into.

PR #79, opened 2023-07-26
Simplify setup

### LINKED ISSUE #33: Validate Snowflake fix for MANAGE WAREHOUSES and revert work around
Right now, the user setup script has a loop that grants the app visibility into all warehouses. This is because the manage warehouses perm was partially broken. Last we checked, it didn't work for viewing queries in the query_history table function and didn't allow query cancellation. We just heard that Snowflake version 7.22, released late last week should have addressed the issue. We should evaluate whether the fix actually works as intended. If so, we should remove the loop from the grant script.

Relevant code: 
https://github.com/sundeck-io/OpsCenter/blob/main/app/ui/setup.py#L88

rymurr:
This removes the first setup section and replaces it with a call to the Snowflake permissions SDK. Once the permissions are granted then the FINALIZE_SETUP script is run

closes #33


app/ui/setup.py
```diff
@@ -64,36 +64,27 @@ def setup_block():
...
 
+    privileges = [
+        "EXECUTE MANAGED TASK",
+        "EXECUTE TASK",
+        "MANAGE WAREHOUSES",
+        "IMPORTED PRIVILEGES ON SNOWFLAKE DB",
+    ]
+    missing_privileges = perms.get_missing_account_privileges(privileges)
+    if len(missing_privileges) > 0:
+        perms.request_account_privileges(missing_privileges)
```
joshelser, line 75: This call is loading the permissions modal element and when the user either exits that modal (by granting permission or cancelling the window), will the `Setup` page be rendered again?

I am wondering if we need to have some call/button to "refresh" the state. I assume we're reloading the page since you don't have any logic after `request_account_privileges` but wanted to double check.

joshelser: great, thanks.

joshelser:
Vicky is addressing the cypress test failures in #76

vqmarkman:
@rymurr , if you rebase your branch and rerun tests, labels tests should not fail anymore.
as for failing setup tests - I can't reproduce it locally :( 
I would like to see another run after the rebase.

PR #92, opened 2023-07-28
Add qtag conditions

### LINKED ISSUE #86: Add another column for qtag to enriched query history
This column will be a transform of the existing qtag column to something more amenable to use in label conditions.

The curretn form is:
``` json
[
{ "source":"foo", "key":"keyname", "value": "valuename"},
...
]
```

we will transform it to
``` json
{ "source": {"key": "value", ...}, ...}
```
using a javascript UDF so that we can apply conditions like qtag, qtag_exists, qtag_matches as found in [sundeck docs](https://docs.sundeck.io/reference/conditions/qtag/)

We will add this as a new field to enriched query history rather than to the base query history to avoid extra migration effort for that table

rymurr:
This PR adds a new column to enriched_query_history which is a transform of the qtag column. The transformed column makes it much easier to write qtag conditions for labels.

We also add a series of functions for manipulating qtags into the `tools` schema

closes #86


rymurr:
I am going to make this a draft till we understand better how to do table migration

bootstrap/016_qtag_helpers.sql
```diff
@@ -0,0 +1,95 @@
...
+} catch {
+return undefined;
+}
+$$;
+-- match a qtag value with regex and with a specific key and source, given the transformed qtag
+create or replace function tools.qtag_matches(qtag variant, source varchar, key varchar, pattern varchar, parameters varchar)
+returns boolean
+as
+$$
+coalesce(rlike(tools.qtag_value(qtag, source, key), pattern, parameters), FALSE)
```
joshelser, line 87: If `tools.qtag_value` returns `undefined`, will `rlike` barf?

if so, could we do something like `coalesce(tools.qtag_value(qtag, source, key), '')` to guard against that?

bootstrap/016_qtag_helpers.sql
```diff
@@ -0,0 +1,95 @@
...
+try {
+  QTAGS.forEach((qtag) => {
+    if (!(qtag['SOURCE'] in ret)) {
+      ret[qtag['SOURCE']] = {};
+    }
+    ret[qtag['SOURCE']][qtag.KEY] = qtag.VALUE
+  });
+  return ret;
+} catch {
+ return undefined;
```
joshelser, line 18: Would `return {}` be less onerous to deal with from the caller's side?

joshelser (dismissed):
lgtm. a few questions about handling edge-cases but i'm not sure how quickly we'd run into edge cases (e.g. malformed qtag comments)

bootstrap/016_qtag_helpers.sql
rymurr: Most (all) of snowflakes functions handle `NULL` as an input and return `NULL` as an output. I think the default handling here is easier than checking for an empty variant

rymurr:
only rebased a conflict. no changes so merging this.

bootstrap/016_qtag_helpers.sql
joshelser: > Most (all) of snowflakes functions handle NULL as an input and return NULL as an output

Cool! That's good to know. I wasn't sure how the js->sql interaction worked in this case (and 30s of perusing of docs didn't lead me to the answer). Agree with your approach.

Issue #94, opened 2023-07-28
External Functions Forbidden access with Snowflake SSO token


anshuldata:
External function definition should be initialised with received token and correct ef_url

anshuldata:
Fixed with [this](https://github.com/sundeck-io/OpsCenter/pull/95) PR

PR #95, opened 2023-07-28
Fix ef definition on receival of Snowflake SSO Token (#94)

anshuldata:


app/ui/setup.py
```diff
@@ -135,7 +135,10 @@ def sundeck_signup_with_email(account, user, region, db):
         msg = st.empty()
         msg.warning("Connecting. Please do not navigate away from this page.")
         token, url = decode_token(token_input)
-        connection.Connection.get().call("INTERNAL.SETUP_SUNDECK_TOKEN", url, token)
+        sndk_token = f"sndk_{token}"
```
rymurr, line 138: how come we have to add `sndk_` here? I thought the api gateway ignored the prefix too?

anshuldata: Yes, even not adding prefix will also work since API gateway ignores it. I choose to prefixed it to keep behavior same as existing code  [here](https://github.com/sundeck-io/OpsCenter/blob/a687a61b3c24a1aa4f399e54a8d1a77c517b714c/bootstrap/021_external_functions.sql#L182).

rymurr: ah, got ut. I didn't know hte gateways still did taht. Prob could be cleaned up but definitely not worth the effort right now

rymurr (approved):

PR #97, opened 2023-07-31
Unit tests for stored procedures example.

vqmarkman:


tests/README.md
```diff
@@ -0,0 +1,27 @@
+# OpsCenter SQL tests
+
+## Running tests
+
+Place the `parameters.py` file in the `tests/sql` directory, with the connection information:
```
rymurr, line 5: It would be better to use the same technique as `deploy/deploy.py` read ~/.snowsql/config

tests/sql/common_utils.py
```diff
@@ -0,0 +1,33 @@
+import uuid
```
rymurr, line 1: I think it may be more idoomatic for these to live in `conftest.py` then they will be available everywhere in pytest

tests/README.md
vqmarkman: agree and will fix.

vqmarkman:
cc @jinfengni - I added test for opscenter-26 and it will become part of the new regression suite. Just FYI.

vqmarkman:
Restructured tests in a bit different way and simplified some things.
New PR is here: https://github.com/sundeck-io/OpsCenter/pull/138

PR #117, opened 2023-08-01
OPSCENTER-116: Fix for multiple app in same account to have different…

anshuldata:
… notification endpoint

	* Each notification endpoint security integration and app name should be different


rymurr:
@anshuldata what will the integration name look like in the web browser?

rymurr:
Is that what the Web Browser will say on the snowflake login screen?

anshuldata:
> Is that what the Web Browser will say on the snowflake login screen?

Fixed Security integration used for auth of notification endpoint to be "Sundeck" in most common cases

Issue #100, opened 2023-07-31
Remove use of `*` in projections


rymurr:
We currently use `*` in projections for the enriched query data. We need to remove this to ensure all tables/views are well defined and to allow us to make improvements/changes to the underlying tables without breaking intermediate stages

PR #134, opened 2023-08-07
OPSCenter-101: Make migration support add or removal column

### LINKED ISSUE #101: Improve migration script
Currently `internal.migrate_if_necessary` only works for additions to the base tables. eg if Snowflake adds fields to the `query_history` table. We have to make it more flexible to be able to support:
* removing of columns from base tables
* additions of columns by us (eg #88 ) 

This is especially tricky right now as we use `*` in projections in a number of places, therefore making the ordinal and total number of columns in these tables important. Before this issue is worked on we therefore need to close #100

Once that is done we should be able to update the migration scripts to simply add columns w/o regard to order and be able to easily remove columns if required. We should also be able ot add columns of our own (eg #88 or #86)

jinfengni:
close #101


rymurr:
@jinfengni I thikn we have to be able to handle cases 1&3. The current mechanism handles case 1 so not handling would be a regression. Case 3 has already broken opscenter for us last month. I think breaking opscenter till we can get a release out is not acceptable. (especially as the removal/addition of columns can happen based on region or specific snowflake account setup and we may not be able to release targeted fixes).

jinfengni:
@rymurr ， probably I did not explain clearly.  

Case1 is not a regression.  With the change for issue #100,  select * is replaced with an explicitly column list. When Snowflake adds a new column to and Sundeck does not add it to the column list,  this new column is just invisible to the view/table.  If Sundeck chooses to add the new column to the column list, it becomes Case2. 

Similar for Case 3.  When Snowflake removes an existing column from `query_history`, which our view DDL uses it,  the view DDL will hit SQL error if we do not explicitly remove that column in our view DDL.  If we choose to remove that column, it becomes Case 4. 

For case 1 and case 3,  since we use the **explicit** column list, in stead of *, I thought we have to maintain the column list ourself.  Did I miss something here?

rymurr:
@jinfengni is there no way to alter the views automatically w/o a release? eg alter the column list and `create or replace` the views? I am concerned that having to wait for a relase from us is not going to be acceptable for the user especially if their app is broken in the meantime.

jinfengni:
@rymurr 

If we want to alter view automatically, one choice is to use select * again.  It will ensure that our view `QUERY_HISTORY_COMPLETE_AND_DAILY` will keep sync with the columns in snowflake's `query_history` atomically. The new migrate script does not require we maintain the ordinal position, so it's ok to use select * in view DDL, IMHO.

rymurr:
If you are happy with that then I am!

jinfengni:
sounds good. I'll give it a try to use select *, and test the new migrate scripts will allow select * in the view DDL.

jinfengni:
Looks like `select *` will not work either.   Snowflake will record the column list when view is created with `select *`. And at view execution, it will compare the column list recorded with the ad-hoc column list, and fail the query if they are out of sync.  

If I have a `query_history_test` with 3 columns, and `sundeck_query_history` use `select *`
```
create view query_history_test as select 'col1' as col1, 'col2' as col2, 'col3' as col3;

create view sundeck_query_history as select * from query_history_test;
```

If two columns are added to `query_history_test` and we do not re-create `sundeck_query_history`,  it would hit error to query `sundeck_query_history`.  

```
select * from sundeck_query_history;
002057 (42601): SQL compilation error:
View definition for 'JINFENG_TESTDB.PUBLIC.SUNDECK_QUERY_HISTORY' declared 3 column(s), but view query produces 5 column(s).
```

So, `select *` will not work, and we have to re-create the view in case Snowflake adds/removes column to/from `query_history`.

jinfengni:
I saw `INTERNAL_REPORTING.QUERY_HISTORY_COMPLETE_AND_DAILY` is created in `bootstrap`.   One choice to re-create the view `INTERNAL_REPORTING.QUERY_HISTORY_COMPLETE_AND_DAILY` is in `migrate_if_necessary`.  But `migrate_if_necessary` is called in `refresh_queries`, still may not happen immediately when Snowflake addes/removes a column to/from `query_history`.

rymurr:
> I saw `INTERNAL_REPORTING.QUERY_HISTORY_COMPLETE_AND_DAILY` is created in `bootstrap`. One choice to re-create the view `INTERNAL_REPORTING.QUERY_HISTORY_COMPLETE_AND_DAILY` is in `migrate_if_necessary`. But `migrate_if_necessary` is called in `refresh_queries`, still may not happen immediately when Snowflake addes/removes a column to/from `query_history`.

Yeah, I think taht is unavoidable. The migratE_if_neccessary happens hourly and thats probably the best we can do

jinfengni:
@rymurr 

I revert the commit of #100, and use `select *` in the view DDL.  In addition, I put the view create DDL inside a stored procedure.  The views that depends on `account_usage.query_history` will be re-created in sp `internal.migrate_view`. 

With such changes,  we can support the case that Snowflake adds a new column, or drops an existing column in `query_history`, atomically, by running `internal.refresh_warehouse_events(true);` and `internal.refresh_queries(true);`.

Two cases are tested, and verified with expected results:

1.  Add a new column `ADDED_COLUMN` to `account_usage.query_history`.
2.  Drop existing column `ROLE_TYPE` from `account_usage.query_history`, and add another new column `ADDED_COLUMN2`.

In both cases, `internal.migrate_view` and `internal.migrate_if_necessary()` work as expected. 

Please take another look, and let me know if there is any issue.

bootstrap/005_table_migrations.sql
```diff
@@ -25,35 +25,62 @@ BEGIN
...
 END;
+
+
+CREATE OR REPLACE PROCEDURE internal.migrate_view()
+    RETURNS STRING
+    LANGUAGE SQL
+    COMMENT = 'Re-create view used in the App. This is required when Snowflake adds a new column to query_history, or removes an existing column from query_history'
+    AS
+BEGIN
+    call INTERNAL.create_view_QUERY_HISTORY_COMPLETE_AND_DAILY();
```
rymurr, line 80: where are these procs defined?

bootstrap/005_table_migrations.sql
```diff
@@ -25,35 +25,62 @@ BEGIN
...
+    call INTERNAL.create_view_enriched_query_history();
+    call INTERNAL.create_view_enriched_query_history_daily();
+    call INTERNAL.create_view_enriched_query_history_hourly();
+    call INTERNAL.create_view_LABELED_QUERY_HISTORY();
```
rymurr, line 84: Does this take into account existing label columns? Or do we have to re run labellign also?

bootstrap/004_labels.sql
```diff
@@ -55,83 +55,7 @@ BEGIN
...
-        QUERY_LOAD_PERCENT,
-        IS_CLIENT_GENERATED_STATEMENT,
-        QUERY_ACCELERATION_BYTES_SCANNED,
-        QUERY_ACCELERATION_PARTITIONS_SCANNED,
-        QUERY_ACCELERATION_UPPER_LIMIT_SCALE_FACTOR,
-        TRANSACTION_ID,
-        CHILD_QUERIES_WAIT_TIME,
-        ROLE_TYPE,
-    $$;
+SELECT *,$$;
```
rymurr, line 58: I thikn we want to keep the change to remove `select *` unless its absolutely crucial to this migration working. We have had nothing but pain from using select * in past iterations of native app and it doesn't look like it is critical for the migration scheme

bootstrap/018_warehouse_updates.sql
```diff
@@ -11,6 +11,7 @@ BEGIN
     let migrate2 string := null;
     if (migrate) then
         SYSTEM$LOG_TRACE('Migrating data.');
+        call internal.migrate_view();
```
rymurr, line 14: the ordering of this looks strange to me. And doesnt it mean that all views will get updated when only the wrehouse tables will get updated in this case?

bootstrap/005_table_migrations.sql
jinfengni: Good point. I should call `UPDATE_LABEL_VIEW()` which will also re-create `Labeled_query_history` view.

bootstrap/004_labels.sql
jinfengni: If we put the column list explicitly in view DDL, I do not see how we can atomically add or drop one column from the view DDL.  One idea is to generate the view DDL by checking the information_schema. However, the generation of view DDL works similar to the auto expansion of `*`, except that it adds more complexities. Sounds to me `select *` or expanding from information_schema is same.

bootstrap/018_warehouse_updates.sql
jinfengni: for this `018_warehouse_updates.sql`, we do not have to call `internal.migrate_view()`, as the view in `018_warehouse_updates.sql` do not uses `query_history` directly.  Probabaly I should have remove this line of code change.

bootstrap/004_labels.sql
jinfengni: If we use `select *` in view DDL,  Snowflake will record the column list expanded from * at the moment view is created.   If the source of `*` is updated, query the  view will hit error. The only way to fix is to re-create the view again.

Here is the example I tried:

- Create a `qh` view with `Col1, Col2`
```
create view qh as select 'col1' as COL1, 'COL2' as COL2;
```

- Create a view `myview`
```
create view myview as select * from qh;
```

- Get the ddl of `myview`. We can see that Col1, Col2 are expanded from * in the view DDL
```
elect get_ddl('view', 'myview');
+--------------------------------+
| GET_DDL('VIEW', 'MYVIEW')      |
|--------------------------------|
| create or replace view MYVIEW( |
| 	COL1,                                                                 |
| 	COL2                                                                 |
| ) as select * from qh;         |
+--------------------------------+
```

-- Change the column list in `qh` from `Col1, Col2`, to `Col3, Col4`.
```
create or replace view qh as select 'col3' as COL3, 'COL2' as COL4;
```

-- Query `myview` hit error
```
select * from myview;
002334 (42601): SQL compilation error: View columns mismatch with view definition for view 'MYVIEW' at line 1, position 14, please re-create the view
```

From this example, `select *` issue can be fixed by re-create the view.  That's what we do in the migration script.  For all the views that depend on `query_history`, we'll re-create those views.

bootstrap/018_warehouse_updates.sql
jinfengni: changed `018_warehouse_updates.sql` such tat we do not call `internal.migrate_view()`.

bootstrap/005_table_migrations.sql
rymurr: ah, i see it now. Does this work ok in a fresh install of opscenter? We are calling stored procs from `005` which arent defined till `016`.

Also, why don't we have to do the same for the warehouses tables?

bootstrap/004_labels.sql
jinfengni: right.  We re-create the views in `refresh_queries`, regardless the underlying is changed or not.  The re-creating views will cause the view to have a new expanded column list from `select *` in case underlying is changed. 

I think the column expansion from `select *` in view creating time is same as the column expansion if we choose to construct the view DDL dynamically from information_schema.

bootstrap/005_table_migrations.sql
jinfengni: I did both `deploy/deploy.py` and `deploy/devdeploy.py`, and  both cases run fine.

For warehouse tables, we use `select * from INTERNAL_REPORTING.CLUSTER_AND_WAREHOUSE_SESSIONS_COMPLETE_AND_DAILY ` when create table.    The view  `INTERNAL_REPORTING.CLUSTER_AND_WAREHOUSE_SESSIONS_COMPLETE_AND_DAILY`  uses explicit column list. If the underlying `ACCOUNT_USAGE.WAREHOUSE_EVENTS_HISTORY` is changed, we do not have to re-create the tables.

rymurr: but we have the same problem as query history? if they remove a column from the table and we haven't released opscenter w/ the changes then the app will be broken?

It seems like this migration pattern is generic for all table types we rely on

jinfengni: I saw we have the following select clause over `warehouse_events_history`.  If the column explicitly referred in the expressions are dropped by Snowflake, we have to change the view DDL in OpsCenter, and redeploy with the modified view DDL.  If Snowflake adds a new column, or removes a column that is not referred in the expression here, then OpsCenter will work as it is without making any migration/change.

In other words, if the columns explicitly used here are dropped by Snowflake, I feel we have to update OpsCenter app, and redeploy it. 

```
    SELECT
        timestamp as filterts,
        timestamp,
        warehouse_id,
        warehouse_name,
        cluster_number,
        case
            when event_name in ('RESUME_WAREHOUSE','SUSPEND_WAREHOUSE') then 'WH'
            when event_name in ('RESUME_CLUSTER','SUSPEND_CLUSTER') then 'CL'
        END as SESSION_TYPE,
        case
            when event_name = 'RESUME_WAREHOUSE' and event_state = 'STARTED' then 1
            when event_name = 'RESUME_CLUSTER' and event_state = 'COMPLETED' then 1
            when event_name = 'SUSPEND_CLUSTER' and event_state = 'COMPLETED' then 2
            when event_name = 'SUSPEND_WAREHOUSE' and event_state = 'COMPLETED' then 2
            else null
        end as SESSION_BEHAVIOR,
        case
            when event_name = 'RESUME_WAREHOUSE' and event_state = 'STARTED' then 1
            when event_name = 'RESUME_CLUSTER' and event_state = 'COMPLETED' then 2
            when event_name = 'SUSPEND_CLUSTER' and event_state = 'COMPLETED' then 3
            when event_name = 'SUSPEND_WAREHOUSE' and event_state = 'COMPLETED' then 4
            else null
        end as EVENT
    FROM ACCOUNT_USAGE.WAREHOUSE_EVENTS_HISTORY
```

jinfengni: > did you drop the db before running teh above?

I did not drop the db before the above `deploy/deploy.py`, and `deploy/devdeploy.py`.   I just re-deploy after dropping the App and dbs, and they run fine.

```
drop application jinfeng_opscenter_app;
drop database jinfeng_opscenter_pkg;

drop database jinfeng_opscenter;

poetry run python deploy/deploy.py -p jinfeng_local -v V1

poetry run python deploy/devdeploy.py -p local_dev
```

rymurr (approved):

PR #138, opened 2023-08-08
Unit test framework for SQL + labels unit tests.

### LINKED ISSUE #30: Add Unit Testing Framework for SQL operations
One of the problems with OpsCenter right now is that it is easy to break things since there are very few tests beyond the cypress tests. We need to look at how we enable more unit testing.

vqmarkman:
Closes #30.

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -6,11 +6,21 @@ def generate_unique_name(prefix, timestamp_string) -> str:
     return name
 
 
-def run_proc(conn, sql) -> str:
+def _get_single_value(conn, sql):
     with conn() as cnx:
         cur = cnx.cursor()
         result = cur.execute(sql).fetchone()
-        return result
+        return result[0]
+
+
+def run_proc(conn, sql) -> str:
+    assert isinstance(result := _get_single_value(conn, sql), (str, type(None)))
+    return result
+
+
+def row_count(conn, sql) -> int:
+    assert isinstance(result := _get_single_value(conn, sql), int)
+    return result
 
 
 def delete_list_of_labels(conn, sql):
@@ -23,11 +33,3 @@ def delete_list_of_labels(conn, sql):
             assert "done" in str(
                 run_proc(conn, delete_label_statement)
             ), "Stored procedure output does not match expected result!"
-
-
-# Used for validating result correctness
-def validate_row_count(conn, sql) -> int:
-    with conn() as cnx:
-        cur = cnx.cursor()
-        result = cur.execute(sql).fetchone()
-        return result
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -6,28 +6,28 @@ def generate_unique_name(prefix, timestamp_string) -> str:
     return name
 
 
-def run_proc(conn_cnx, sql) -> str:
-    with conn_cnx() as cnx:
+def run_proc(conn, sql) -> str:
+    with conn() as cnx:
         cur = cnx.cursor()
         result = cur.execute(sql).fetchone()
         return result
 
 
-def delete_list_of_labels(conn_cnx, sql):
+def delete_list_of_labels(conn, sql):
     print(f"[INFO] SQL in delete function: {sql}")
 
-    with conn_cnx() as cnx:
+    with conn() as cnx:
         cur = cnx.cursor()
         for name in cur.execute(sql).fetchall():
             delete_label_statement = f"call ADMIN.DELETE_LABEL('{name[0]}');"
             assert "done" in str(
-                run_proc(conn_cnx, delete_label_statement)
+                run_proc(conn, delete_label_statement)
             ), "Stored procedure output does not match expected result!"
 
 
 # Used for validating result correctness
-def validate_row_count(conn_cnx, sql) -> int:
-    with conn_cnx() as cnx:
+def validate_row_count(conn, sql) -> int:
+    with conn() as cnx:
         cur = cnx.cursor()
         result = cur.execute(sql).fetchone()
         return result
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -1,0 +1,33 @@
+import uuid
+
+
+def generate_unique_name(prefix, timestamp_string) -> str:
+    name = prefix + "_" + str(uuid.uuid4()) + "_" + timestamp_string
+    return name
+
+
+def run_proc(conn_cnx, sql) -> str:
+    with conn_cnx() as cnx:
+        cur = cnx.cursor()
+        result = cur.execute(sql).fetchone()
+        return result
+
+
+def delete_list_of_labels(conn_cnx, sql):
+    print(f"[INFO] SQL in delete function: {sql}")
+
+    with conn_cnx() as cnx:
+        cur = cnx.cursor()
+        for name in cur.execute(sql).fetchall():
+            delete_label_statement = f"call ADMIN.DELETE_LABEL('{name[0]}');"
+            assert "done" in str(
+                run_proc(conn_cnx, delete_label_statement)
+            ), "Stored procedure output does not match expected result!"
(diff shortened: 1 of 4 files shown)


test/unit/test_label_procs.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+# List of test cases with statements and expected error messages
+test_cases = [
+    (
+        "call ADMIN.CREATE_LABEL(NULL, NULL, NULL, 'compilation_time > 5000');",
+        "Name must not be null.",
+    ),
+    (
+        "call ADMIN.CREATE_LABEL('{label}', NULL, 100, 'compilation_time > 5000');",
```
joshelser, line 81: Is an empty label name (`''`) allowed or disallowed?

test/unit/test_label_procs.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def test_smoke_create_drop_label(conn_cnx, timestamp_string):
+    label = generate_unique_name("label", timestamp_string)
+    sql = f"call ADMIN.CREATE_LABEL('{label}', NULL, NULL, 'rows_produced > 100');"
+
+    # create_label returns NULL in case of successful label creation
+    assert (
+        run_proc(conn_cnx, sql)[0] is None
+    ), "Stored procedure did not return NULL value!"
```
joshelser, line 16: nit: the parens around the `assert` condition are generally unnecessary (here and below)

test/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+
+    cnx = create_connection(profile, **kwargs)
+    try:
+        yield cnx
+    finally:
+        cnx.close()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def conn_cnx():
```
joshelser, line 108: This name is a little duplicative? Maybe just `def conn()`?

test/common_utils.py
```diff
@@ -0,0 +1,33 @@
...
+def delete_list_of_labels(conn_cnx, sql):
+    print(f"[INFO] SQL in delete function: {sql}")
+
+    with conn_cnx() as cnx:
+        cur = cnx.cursor()
+        for name in cur.execute(sql).fetchall():
+            delete_label_statement = f"call ADMIN.DELETE_LABEL('{name[0]}');"
+            assert "done" in str(
+                run_proc(conn_cnx, delete_label_statement)
+            ), "Stored procedure output does not match expected result!"
```
joshelser, line 25: This is a little wonky by mixing both test cleanup logic with OpsCenter "application" logic.

What about making API methods that could do things like create, list, or delete labels (calling the appropriate SQL in those methods), and then repurpose this method to just be a "teardown" method which calls those new methods? e.g.

```
def teardown(conn_cnx):
    ...
    conn = conn_cnx()
    for name in list_labels(conn):
        delete_label(conn, name)
```

I think this will have the bonus of making your test code much more terse (less SQL) and easier to extend.

joshelser (commented):
I left a comment in test/common_utils.py about taking the CRUD operations against labels and making them be "test API" (calling the procedures in the ADMIN schema).

I think that suggestion would clean up a lot of the test methods:
* don't need to create as much sql in the test method
* don't need to parse the SQL rows in test logic

wdyt?

test/common_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+# Used for validating result correctness
+def validate_row_count(conn_cnx, sql) -> int:
```
rymurr, line 29: this method is identical to `run_proc` why have both?

test/conftest.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Returns value of --profile argument
+def snowsql_profile_name() -> str:
+    for i, arg in enumerate(sys.argv):
+        if arg == "--profile" and i + 1 < len(sys.argv):
+            return sys.argv[i + 1]
+    return "opscenter"
+
+
+def get_db_parameters_from_snowsql_config(profile):
```
rymurr, line 29: whynot use `deploy/helpers.py` for this?

test/unit/test_label_procs.py
rymurr: I think to have them on separate lines they are neccessary

vqmarkman: That's python's linter (ruff) doings ... 

```
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
- files were modified by this hook

Found 1 error (1 fixed, 0 remaining).

black....................................................................Failed
- hook id: black
- files were modified by this hook
```


```
     sql = f"call ADMIN.DELETE_LABEL('{label}');"
-    assert run_proc(conn_cnx, sql)[0] == "done", "Stored procedure did not return NULL value!"
+    assert (
+        run_proc(conn_cnx, sql)[0] == "done"
+    ), "Stored procedure did not return NULL value!"
```

test/conftest.py
vqmarkman: Great suggestion, 4 characters less to type :) Committed the change.

test/common_utils.py
vqmarkman: They return different data types and I felt it would be bad form to combine them even though they are super simple ...
Plus for readability/usability: it's harder to make a mistake when you have very different function names.
If you feel strongly about it, I will make a change.

test/conftest.py
vqmarkman: doh ... I don't know why I did not do it in the first place. Fixed. Thanks for pointing it out.

test/unit/test_label_procs.py
vqmarkman: great catch! Added test case.

test/common_utils.py
vqmarkman: Josh, the reason I'm querying internal table for labels is because I'm protecting against the case where due to a bug of some sort, negative test did not fail as expected and label got created. In this case, I'm finding every label that was created in a test session and deleting it as there is no other method to "list labels" (or I'm not aware of it)

As to "less SQL" code ... Ryan and I wanted to make this as simple as possible: when test fails, you see exactly which SQL statement failed without any extra debugging + easy to add specifically SQL tests... Dunno ... let's talk. Maybe I'm missing something important. Thanks for the review, I found it very helpful!

test/common_utils.py
rymurr: its just code duplication it doesn't seem useful to me. We are not doing type checks anywhere so the types are only informational to the user,

if you want to be careful about types and more informational I would do something like

``` python
def _get_single_value(conn_ctx: cursor, sql: str):
    with conn() as cnx:
        cur = cnx.cursor()
        result = cur.execute(sql).fetchone()
        return result

def run_proc(conn_ctx: cursor, sql: str) -> str:
    result = _get_single_value(conn_ctx, sql)
    assert isinstance(result, str)
    return result

def row_count(conn_ctx: cursor, sql: str) -> int:
    result = _get_single_value(conn_ctx, sql)
    assert isinstance(result, int)
    return result
```

vqmarkman: That's elegant! Lemme do it. Thank you.

joshelser (approved):

test/common_utils.py
vqmarkman: done.

joshelser (approved):
lgtm. Do you have plans to also wire this up to a github workflow?

It would also be good to switch over the `print` statements to using a proper logger, but we can do that whenever.

vqmarkman:
> lgtm. Do you have plans to also wire this up to a github workflow?
> 
> It would also be good to switch over the `print` statements to using a proper logger, but we can do that whenever.

oh yes, logger! I will file an issue about it and wiring to a GHA workflow -  yes ... we just need to decide if we want to run these tests with every PR or not.  Two tickets.

PR #144, opened 2023-08-09
Recreaste catalog.labels view

### LINKED ISSUE #139: catalog.probes view was not updated after internal.probes schema was changed
```
snowflake> select * from catalog.probes;
View definition for 'OPSCENTER.CATALOG.PROBES' declared 6 column(s), but view query produces 9 column(s).
```

In support of the #91 work, the `INTERNAL.PROBES` table had some additional columns added to it via a procedure which conditionally altered the table. However, the view was not updated.

Uninstalling and re-installing OpsCenter will fix this as a workaround, but we need to address this in a backwards compatible way.

Maybe related to the work on-going in #134

rymurr:
closes #139

### DIFF
--- a/bootstrap/004_labels.sql
+++ b/bootstrap/004_labels.sql
@@ -13,7 +13,7 @@ BEGIN
 
     -- Recreate the view to avoid number of column mis-match. Should be cheap and only run on install/upgrade, so it's OK
     -- if we run this unnecessarily.
-    CREATE VIEW CATALOG.LABELS IF NOT EXISTS AS SELECT * FROM INTERNAL.LABELS;
+    CREATE OR REPLACE VIEW CATALOG.LABELS AS SELECT * FROM INTERNAL.LABELS;
 EXCEPTION
     WHEN OTHER THEN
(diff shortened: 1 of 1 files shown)


bootstrap/004_labels.sql
```diff
@@ -11,6 +11,9 @@ BEGIN
         UPDATE INTERNAL.LABELS SET LABEL_MODIFIED_AT = CURRENT_TIMESTAMP WHERE LABEL_MODIFIED_AT IS NULL;
     END IF;
 
+    -- Recreate the view to avoid number of column mis-match. Should be cheap and only run on install/upgrade, so it's OK
+    -- if we run this unnecessarily.
+    CREATE VIEW CATALOG.LABELS IF NOT EXISTS AS SELECT * FROM INTERNAL.LABELS;
```
joshelser, line 16: needs to be `create or replace` and not have `if not exists`, else this won't recreate the view, right?

rymurr: fixed

joshelser (approved):

PR #145, opened 2023-08-09
add quotes to security integration

rymurr:


app/ui/setup.py
```diff
@@ -252,7 +252,7 @@ def sundeck_signup_with_snowflake_sso(
 def generate_code_to_create_sundeck_account(
     db: str, sf_region: str, sd_deployment: str
 ) -> str:
-    security_integration_name = "Sundeck"
+    security_integration_name = '"Sundeck"'
     if sd_deployment != "prod":
         security_integration_name = f"""SUNDECK_OAUTH_{db.upper()}"""
```
joshelser, line 257: We would also need quoting for the non-official install if someone is using a database name that requires quoting.

Fine if that's done separately.

joshelser (approved):

app/ui/setup.py
rymurr: `f"""SUNDECK_OAUTH_{db.upper()}"""`
will never need quoting. Its already all uppercase

joshelser: I'm thinking about the case where `db` was `'my database'`. Yes, it would be normalized to `'MY DATABASE'`, but creating the security integration would fail since we do not have it quoted (extra token when the sql is parsed).

This would only affect developers and could be easily addressed by using a database name that doesn't require quoting. That's why I mentioned we could do it later.

rymurr: gotcha, fixed

joshelser (approved):

PR #151, opened 2023-08-10
Do not pass quoted security integration name to SHOW_OAUTH_CLIENT_SEC…

joshelser:
…RETS

Regression from #145

The quoting of the security integration name works fine. However, the SYSTEM$SHOW_OAUTH_CLIENT_SECRETS function cannot handle the double-quotes inside of the integration name.

### DIFF
--- a/app/ui/setup.py
+++ b/app/ui/setup.py
@@ -252,13 +252,19 @@ def sundeck_signup_with_snowflake_sso(
 def generate_code_to_create_sundeck_account(
     db: str, sf_region: str, sd_deployment: str
 ) -> str:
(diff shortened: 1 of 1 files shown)


joshelser:
When enabling snowflake SSO, the generated SQL to run includes

```
let oauth_info variant := (parse_json(SYSTEM$SHOW_OAUTH_CLIENT_SECRETS('"Sundeck"')));
```

This fails with the error `invalid value ["Sundeck"] for parameter 'OAuth Client Integration name'`. Simple fix to just not pass the quoted name to this function.

rymurr (dismissed):

rymurr (approved):

Issue #110, opened 2023-07-31
Migration for predefined labels and probes


rymurr:
We will add/modify predefined labels only when none of them have been modified by the user.

We will perform a join on the labels table and the predefined label table (#108) on the condition text and name. If the label modified date (#109) is greater than the predefined label created date or if a predefined label is missing (eg deleted) then we will skip updating the labels. Otherweise we will upgrade all label definitions.

Same process for probes.

jinfengni:
@rymurr , I notice that `labels` has the column `LABEL_CREATED_AT`, in addition to `LABEL_MODIFIED_AT`.  However, `probes` table has only `PROBE_MODIFIED_AT`.   Do we have to add `PROBE_CREATED_AT` column to `PROBES` table too?

```
call admin.run_as_app('desc table  internal.probes');
+----------------------+-------------------+--------+-------+---------+-------------+------------+-------+------------+---------+-------------+
| name                 | type              | kind   | null? | default | primary key | unique key | check | expression | comment | policy name |
|----------------------+-------------------+--------+-------+---------+-------------+------------+-------+------------+---------+-------------|
| NAME                 | VARCHAR(16777216) | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| CONDITION            | VARCHAR(16777216) | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| NOTIFY_WRITER        | BOOLEAN           | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| NOTIFY_WRITER_METHOD | VARCHAR(16777216) | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| NOTIFY_OTHER         | VARCHAR(16777216) | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| NOTIFY_OTHER_METHOD  | VARCHAR(16777216) | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| CANCEL               | BOOLEAN           | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
| ENABLED              | BOOLEAN           | COLUMN | Y     | NULL    | N           | N          | NULL  | NULL       | NULL    | NULL        |
[...]

Issue #109, opened 2023-07-31
Add modified at column to labels and probes tables


rymurr:
This column will be set to the created date at creation and to the current timestamp on any later modificaiton. Primary use for this is migrating new predefined labels and probes (#108 )

PR #172, opened 2023-08-18
OPSCENTER-27: add predefined labels

### LINKED ISSUE #27: Add a selection of useful default labels
Right now the OpsCenter install includes no default labels. We should update such that user has a bunch of useful labels immediately.

jinfengni:
close #27


bootstrap/004_labels.sql
```diff
@@ -134,6 +134,59 @@ BEGIN
 
 END;
 
+CREATE OR REPLACE PROCEDURE INTERNAL.CREATE_PREDEFINED_LABEL(name text, grp text, rank number, condition text)
```
rymurr, line 137: I dont think we need this for the predefined labels. We should just be able to insert direclty in the other proc.

rymurr (commented):
LGTM, 1 minor comment to address.

Out of curiosity, will we define the migration and population of hte `LABELS` table in a separate PR?

bootstrap/004_labels.sql
jinfengni: Earlier one, I used direct "insert into" to populate the predefined labels.  I realized that using a stored proc could apply the same checks over "name", "condition" or other properties in the same way as the regular labels. That's the main reason I choose to use a stored procedure to 1) check the input, 2) insert into table.

jinfengni:
> will we define the migration and population of hte LABELS table in a separate PR
The migration will be in a separate PR.  The population of LABELS table will be in this PR (I'm currently working on it).

jinfengni:
I added another SP `MIGRATE_PREDEFINED_LABELS`, which will both populate and migrate the predefined labels to `labels` tables. 

1. In the first run of deploy, when `labels` does not have the predefined labels, it will populate those predefined labels to `labels` table.
2. On subsequent run of deploy, it will migrate the predefined labels to `labels` table.

jinfengni:
@rymurr I choose to use the following stored procedures :
- `INTERNAL.POPULATE_PREDEFINED_LABELS()` :  populate `predefined_labels` table, 
- `INTERNAL.MIGRATE_PREDEFINED_LABELS` : populate `labels` from `predefined_labels` if the predefined_label does not exist, or migrate modified `predefined_label` to `labels` table.

Both the two stored procedures are called in `090_postsetup.sql`.   

- `INTERNAL.POPULATE_PREDEFINED_LABELS()` 
It calls `INTERNAL.UPSERT_PREDEFINED_LABEL` to either insert or update the predefined labels in `predefined_labels` table.  When a predefined label is first inserted, both `LABEL_CREATED_AT` and `LABEL_MODIFIED_AT` are set to be current_timestamp().  When an existing predefined label is modified upon next call of  `INTERNAL.POPULATE_PREDEFINED_LABELS()`,  only `LABEL_MODIFIED_AT` is updated. 

- `INTERNAL.MIGRATE_PREDEFINED_LABELS`
It uses a `Merge` statement to handle both insertion or migration of `predefined_labels` to `labels`.  We condition `labels`.`LABEL_MODIFIED_AT` == `predefined_labels`.`LABEL_CREATED_AT` to indicate that the predefined label in `labels` tables is sync with the one in `predefined_labels` table. 

Migration:  When name/condition are same, `labels`.`LABEL_MODIFIED_AT` <= `predefined_labels`.`LABEL_CREATED_AT`,  we upgrade the predefined label in `labels` table, and set `labels`.`LABEL_MODIFIED_AT` to be same as `predefined_labels`.`LABEL_CREATED_AT`, meaning they are sync.  If users modifies the `predefined_label` in `labels` column,  such condition will not met, and we will not update the predefined label in `labels` table. 

Populating into `labels` table :  When no row with same name/condition is found in `labels` table, we insert such predefined label into `labels`. 

The above processing will not be able to different the case where the predefined label does not exist in `labels` table, and the case where the predefined label is deleted by the user.  In both case, when `MIGRATE_PREDEFINED_LABELS` is called, we will insert a new row into `labels` for such predefined label.  This is different from the requirement in issue #110.


> If the label modified date (https://github.com/sundeck-io/OpsCenter/issues/109) is greater than the predefined label created date or if a predefined label is missing (eg deleted) then we will skip updating the labels. Otherweise we will upgrade all label definitions.

Please let me know your comments. Thanks!

jinfengni:
btw: in the current code, I only insert two predefined labels.  If the above logic makes sense, I'll insert more predefined lables from 1p.

jinfengni:
I added another `INITIALIZE_LABELS` stored procedure to do the one-time initialization of `labels` table from `predefined_labels` table, when 
- it has not done before 
- there is no user created label before

This is based on the specification from 1p (Thanks to @vqmarkman )

> We will add a row to internal.config to state that pre-defined labels are up to v1. We will not make any changes to labels or probes if that row is present

bootstrap/000_setup_base_tools.sql
```diff
@@ -16,6 +16,23 @@ BEGIN
       INSERT (key, value) VALUES (source.key, source.value);
 END;
 
+CREATE OR REPLACE PROCEDURE internal.get_config(key string) RETURNS STRING LANGUAGE SQL
+    AS
+DECLARE
+    EXCEPTION_DUP_KEYS EXCEPTION(-20001, 'More than 1 rows in table internal.config for the specified key');
```
rymurr, line 22: wouldnt it be easier to use a unique constraint on the key? That would simplify this method considerably

bootstrap/004_labels.sql
```diff
@@ -134,6 +134,102 @@ BEGIN
 
 END;
 
+CREATE OR REPLACE PROCEDURE INTERNAL.UPSERT_PREDEFINED_LABEL(name text, grp text, rank number, condition text)
+    RETURNS text
+    LANGUAGE SQL
+    EXECUTE AS OWNER
+AS
+BEGIN
+    if (:name is null) then
```
rymurr, line 143: I don't think we need to do validation for predefined labels. Its not user input and we cant do anything if it fails, installation or upgrade would fail.

Instead we should just test (in `test_label_procs.py` ) that the conditions are valid at commit time (the only time we can actually do anything about the conditions being bad).

bootstrap/004_labels.sql
```diff
@@ -134,6 +134,102 @@ BEGIN
...
+    WHEN NOT MATCHED THEN
+    INSERT
+        ("NAME", "GROUP_NAME", "GROUP_RANK", "LABEL_CREATED_AT", "CONDITION", "LABEL_MODIFIED_AT")
+        VALUES (s.name, s.grp, s.rank,  current_timestamp(), :condition, current_timestamp());
+    outcome := null;
+
+    return outcome;
+END;
+
+CREATE OR REPLACE PROCEDURE INTERNAL.DELETE_PREDEFINED_LABEL(name text)
```
rymurr, line 179: predefined_lables is a immutable table, I don't think deleting things from it makes sense

bootstrap/004_labels.sql
```diff
@@ -134,6 +134,102 @@ BEGIN
...
+BEGIN
+    if (:name is null) then
+      return 'Name must not be null.';
+    end if;
+
+    DELETE FROM internal.predefined_labels where name = :name;
+    return 'done';
+END;
+
+--CREATE OR REPLACE PROCEDURE INTERNAL.MIGRATE_PREDEFINED_LABELS()
```
rymurr, line 193: please remove the commented proc

bootstrap/004_labels.sql
```diff
@@ -134,6 +134,102 @@ BEGIN
...
+    LANGUAGE SQL
+    EXECUTE AS OWNER
+AS
+BEGIN
+    let labelcnt number := (SELECT COUNT(*) FROM internal.labels);
+
+    let label_inited text := '';
+    label_inited := (CALL INTERNAL.get_config('LABELS_INITED'));
+
+    if (labelcnt > 0 OR label_inited = 'YES') THEN
```
rymurr, line 221: I don't think this is the logic defined in #110 
what happens if the labels were inited by us but never changed and we ship more/improved labels?

bootstrap/004_labels.sql
```diff
@@ -203,3 +299,35 @@ EXCEPTION
...
+
+CREATE OR REPLACE PROCEDURE INTERNAL.POPULATE_PREDEFINED_LABELS()
+    RETURNS text
+    LANGUAGE SQL
+    EXECUTE AS OWNER
+AS
+BEGIN
+    let outcome string := '';
+
+    BEGIN TRANSACTION;
```
rymurr, line 311: I wonder if the combo of a merge into and a cte of all hte labels will be more readable and wouldn't require the explicit transaction. Something like:

``` sql
with foo as (
select * from (values ('Large Results, 'rows_produced > 5000000000'), 'writes', 'query_type in (\'CREATE_TABLE_AS_SELECT\', \'INSERT\')'))
) MERGE INTO internal.predefined_labels t using foo ...
```

test/common_utils.py
```diff
@@ -23,6 +23,11 @@ def row_count(conn, sql) -> int:
     return result
 
 
+def run_sql(conn, sql) -> str:
```
rymurr, line 26: how is this different from `run_proc`?

test/unit/test_label_procs.py
```diff
@@ -182,3 +183,79 @@ def test_create_label_with_empty_string_name(conn, timestamp_string):
     assert "done" in str(
         run_proc(conn, sql)
     ), "Stored procedure output does not match expected result!"
+
```
rymurr, line 186: I think for this test we only really need to validate that the conditions compile. Somethign like:

``` sql
conditions := (select condition from internal.predefined_labels);
for condition in conditions
select condition from query_history where false;
end;
```
the above is pseudo code, we would likely want to report which failed and which passed and not just fail hard on the whole test

bootstrap/000_setup_base_tools.sql
jinfengni: good idea.

bootstrap/004_labels.sql
jinfengni: This `DELETE_PREDEFINED_LABEL` is used for test case. I can remove it.

bootstrap/004_labels.sql
jinfengni: removed.  this commented proc should be part of pr for migration (#110).

bootstrap/004_labels.sql
jinfengni: this logic is for initialization of user labels.

Per the slack https://sundeck-io.slack.com/archives/C03NCM06ARY/p1690243065568569, my understanding is we would do initialization for new install and old install.

New install:
  - labelCnt = 0  and LABELS_INITED is not done ==> we will call this SP

Old install:
  - LABELS_INITED is not done.    Only if there is no user created labels (labelCnt = 0) we will call this SP.

For either new install or old install, once this SP is called,  it will not INSERT new labels again.  When we ship more predefined labels, it would INSERT for new install or old install that met the above condition, but not the install who have called this SP before.

test/common_utils.py
jinfengni: `run_proc` assert the returned result is string type. `run_sql` will convert the returned result to string type, when result is other type.

test/unit/test_label_procs.py
jinfengni: I added a stored procedure. `INTERNAL.VALIDATE_PREDEFINED_LABELS` to do the check.

jinfengni:
@rymurr , thanks for your review comments. I revised the code from the feedback.

I'll have a separate PR for migration of predefined_label (#110).

bootstrap/004_labels.sql
jinfengni: Good suggestion. I used a MERGE statement with a SELECT ... values as the source.

```
    MERGE INTO internal.predefined_labels t
    USING (
        SELECT *
        from (values
                ('Large Results', null, null, 'rows_produced > 50000000'),
                ('Writes', null, null, 'query_type in (\'CREATE_TABLE_AS_SELECT\', \'INSERT\')')
             )) s (name, grp, rank, condition)

```

bootstrap/000_setup_base_tools.sql
```diff
@@ -16,6 +32,19 @@ BEGIN
       INSERT (key, value) VALUES (source.key, source.value);
 END;
 
+CREATE OR REPLACE PROCEDURE internal.get_config(key string) RETURNS STRING LANGUAGE SQL
+    AS
+BEGIN
+    let cnt number := (SELECT COUNT(*) AS cnt FROM internal.config WHERE key = :key);
```
rymurr, line 38: I don't think this check is needed. Snowflake will return `null` if key is not found

rymurr (commented):
one minor nit, otherwise looks great!

bootstrap/000_setup_base_tools.sql
jinfengni: good point. The query would return 0 row when key is not found, but the assignment `let config_value string := (query)` would get `config_value = null`.  Removed the check.

rymurr (approved):

Issue #125, opened 2023-08-02
Create fine grained cost control artifacts in bootstrap


rymurr:
This moves the table/udf/proc definitions from the sundeck predefined flow to opscenter.

- [x] boolean flag (defautls to on) to generate consumption data
- [x] probe task turned on if len(probes) >= 0 or bool above is true
- [x] UDF to send consumption to Sundeck for consumption limit hook
- [x] stored proc to generate consumption data and sedn to udf
- [x]  add stored proc to probe task and if statement o skip if bool is off
- [x] circumvent the query history table function limit of 10k rows per call
- [x] revert https://github.com/sundeck-io/OpsCenter/pull/199 to re-enable after predefined probes & labels are released

joshelser:
I spent a part of day testing this and everything looks good to me. I'll resolve this issue when it gets released.

PR #176, opened 2023-08-18
Compute daily usage in probe task

joshelser:
ref #125

* Adds internal.config row to control quota consumption `generate_consumption`
* Defaults generate_consumption='true'
* uses query_history table function with its limitations for first pass
* create a user-facing view that includes consumption data
* Adds external function to the app and reports the consumption data

non-goals in this PR
* ~~External function to send to sundeck~~
* Handling more than 10k queries/day
* Caching of query history rows


bootstrap/090_post_setup.sql
```diff
@@ -69,80 +69,109 @@ CREATE OR REPLACE TASK TASKS.PROBE_MONITORING
 DECLARE
   sql STRING;
 BEGIN
-    CALL INTERNAL.GET_PROBE_SELECT() into :sql;
```
joshelser, line 72: This was largely changed pushed down into another scripting block to make sure we always do probes and quota consumption. I'll comment on what actually changed.

bootstrap/090_post_setup.sql
```diff
@@ -69,80 +69,109 @@ CREATE OR REPLACE TASK TASKS.PROBE_MONITORING
...
-        let action string := act.ACTION_TAKENS;
-        let query_id string := act.QUERY_ID;
-        insert into internal.probe_actions select CURRENT_TIMESTAMP(), :name, :query_id, parse_json(:action), :outcome;
-    END FOR;
-EXCEPTION
-    WHEN OTHER THEN
-        SYSTEM$LOG_ERROR(OBJECT_CONSTRUCT('error', 'Exception occurred during probe monitoring.', 'outcome', :outcome, 'probe_action' :action, 'SQLCODE', :sqlcode, 'SQLERRM', :sqlerrm, 'SQLSTATE', :sqlstate));
-        insert into internal.probe_actions select CURRENT_TIMESTAMP(), :name, :query_id, parse_json(:action), :outcome;
+    SYSTEM$LOG_DEBUG('probe_monitoring task beginning');
+    let outcome string := '';
```
joshelser, line 73: Lifted `outcome` out of the for loop because the `EXCEPTION` statement tried to reference it. See bigger problem below.

bootstrap/090_post_setup.sql
```diff
@@ -69,80 +69,109 @@ CREATE OR REPLACE TASK TASKS.PROBE_MONITORING
...
+            let name string := act.PROBE_NAME;
+            let action string := act.ACTION_TAKENS;
+            let query_id string := act.QUERY_ID;
+            insert into internal.probe_actions select CURRENT_TIMESTAMP(), :name, :query_id, parse_json(:action), :outcome;
+        END FOR;
+    EXCEPTION
+        WHEN OTHER THEN
+            SYSTEM$LOG_ERROR(OBJECT_CONSTRUCT('error', 'Exception occurred evaluating probes against queries.', 'outcome', :outcome, 'probe_action' :action, 'SQLCODE', :sqlcode, 'SQLERRM', :sqlerrm, 'SQLSTATE', :sqlstate));
+            -- TODO this has been broken as it refers to variables that may not be defined. Fix it.
+            insert into internal.probe_actions select CURRENT_TIMESTAMP(), 'unknown name', 'unknown query id', null::variant, :outcome;
```
joshelser, line 149: I think this has been broken for a while. All of these variables need to be defined outside this block to be usable here.

bootstrap/090_post_setup.sql
```diff
@@ -69,80 +69,109 @@ CREATE OR REPLACE TASK TASKS.PROBE_MONITORING
...
+            -- No RAISE to avoid skipping the consumption check
```
joshelser, line 150: Function change; no `RAISE;` so we always try to do consumption too.

bootstrap/090_post_setup.sql
```diff
@@ -69,80 +69,109 @@ CREATE OR REPLACE TASK TASKS.PROBE_MONITORING
...
+    SYSTEM$LOG_DEBUG('finished probe_monitoring, check quota usage consumption');
+
+    let generate_consumption boolean := (select internal.is_consumption_enabled());
+    if (:generate_consumption) THEN
+        SYSTEM$LOG_INFO('starting usage consumption');
+        let start_time timestamp_ltz := (select current_timestamp());
+        BEGIN
+            call internal.get_daily_quota_select() into :sql;
+            execute immediate sql;
+            let quota_usage object := (select daily_quota from table(result_scan(last_query_id())));
```
joshelser, line 161: I found out why we had this pattern for probes -- can't query the information_schema table functions in an owner's rights procedure. But, querying them from a task _is_ ok.

joshelser:
The data generated by this task looks something like:
```
{
  "role": {
    "ACCOUNTADMIN": 3.337773308,
    "JOSH_OPSCENTER_APP": 0.519999428,
    "JOSH_OPSCENTER_DEBUG": 0.176747884,
    "PUBLIC": 0.243823082,
    "SUNDECK_ADMIN": 0.02360844,
    "SUNDECK_READ": 0.000491
  },
  "user": {
    "SYSTEM": 0.349825784,
    "TEST": 0.228273492,
    "jinfeng@sundeck.io": 0.00040237,
    "josh@sundeck.io": 3.674723498,
    "ryan@sundeck.io": 0.009568968
  }
}
```

bootstrap/090_post_setup.sql
joshelser: I think this is a latent problem that I'm fixing in #178

bootstrap/041_quotas.sql
```diff
@@ -0,0 +1,63 @@
...
+
+-- returns SQL which reads the query_history table function and aggregates the credits used by both user and role. executed within a task
+-- as an owner's rights procedure cannot call the query_history table function.
+create or replace procedure internal.get_daily_quota_select()
+returns string
+language sql
+as
+BEGIN
+    let s string := $$
+-- TODO this recomputes the usage since midnight every time it runs. We should cache this for the day and update quotas based on deltas.
```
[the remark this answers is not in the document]
joshelser: yep, missed it yesterday.

bootstrap/041_quotas.sql
```diff
@@ -0,0 +1,63 @@
...
+-- TODO the table func caps out at 10k rows which could miss queries for busy accounts.
+with todays_queries as(
+    select
+        total_elapsed_time,
+        credits_used_cloud_services,
+        warehouse_size,
+        user_name,
+        role_name
+    from table(information_schema.query_history(RESULT_LIMIT => 10000, END_TIME_RANGE_START => date_trunc('day', current_timestamp()), END_TIME_RANGE_END => current_timestamp()))
+    where end_time > date_trunc('day', current_timestamp())
```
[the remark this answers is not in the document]
joshelser: At some point I convinced myself this was necessary as I was seeing incomplete queries (which were being returned with an `END_DATE` of the unix epoch). However, this might have been me testing without the `END_TIME_RANGE_*` attributes. Let me dbl check and either comment or remove.

bootstrap/041_quotas.sql
```diff
@@ -0,0 +1,63 @@
...
+),
+costed_queries as (
+    select greatest(0,total_elapsed_time)*zeroifnull(wc.credits_per_milli) + credits_used_cloud_services as credits_used, user_name, role_name
+    from todays_queries tq left outer join INTERNAL_REPORTING.WAREHOUSE_CREDITS_PER_SIZE wc on tq.warehouse_size = wc.warehouse_size
```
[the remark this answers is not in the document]
joshelser: Taking your word that it would be more efficient :). Happy to convert it to a udf.

bootstrap/041_quotas.sql
```diff
@@ -0,0 +1,63 @@
...
+
+create or replace function internal.is_consumption_enabled()
+returns boolean
+as
+$$
+    select coalesce((select TRY_TO_BOOLEAN(value) from internal.config where key = 'generate_consumption'), TRUE)
+$$;
+
+-- Create the 'generate_consumption' config value set to TRUE only when it doesn't exist
+MERGE INTO internal.config AS target
```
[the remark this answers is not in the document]
joshelser: I thought when we talked last, you thought it was best that we only have an internal config for doing this computation that can be changed as a break-glass situation, but that we wanted to do this work by default for folks. Am I confusing this configuration with something else?

It's easy to add another checkbox to the "tasks" page if we want to make a public-facing config option.

bootstrap/041_quotas.sql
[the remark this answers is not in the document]
rymurr: https://github.com/sundeck-io/sundeck/blob/main/predefined_flows/defn_custom_hooks/quotas.yaml#L72

joshelser:
I'm doing another pass of manual testing, but I think I've addressed all of your feedback, @rymurr

joshelser:
Talked offline with Ryan briefly about this. We agreed that it would ultimately be cleaner to put this work in its own task rather than on the PROBE_MAINTENANCE task on a 5min frequency.

rymurr (dismissed):

joshelser:
Whoops! Pre-commit found a bug. Forgot to unwind the changes to enable/disable the probe_monitoring task since separating the cost control to its own task.

joshelser:
Two bugs:

* Undo the extra logic to resume/suspend the probe monitoring task (now that cost control is on its own task)
* Auto-start the cost control after creating it

rymurr (approved):

PR #179, opened 2023-08-21
Quick fix to avoid referencing vars that are out of scope

### LINKED ISSUE #178: Probe monitoring task failing
Noticed this when going through release motions. I think it was something missed in an earlier release.

joshelser:
Fixes #178

Manually tested the fix. Not yet sure how to add tests for this task.

### DIFF
--- a/bootstrap/090_post_setup.sql
+++ b/bootstrap/090_post_setup.sql
@@ -141,8 +141,8 @@ Query Text: {bt}{bt}{bt}{query_text}{bt}{bt}{bt}
     END FOR;
 EXCEPTION
     WHEN OTHER THEN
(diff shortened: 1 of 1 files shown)


joshelser:
Chatting with Ryan -- we're both baffled as to why test-deploy has started failing on a grant. The sql being run in deploy.py hasn't changed and this PR shouldn't affect anything to do with the app package.

I'll look at it later (maybe something needs tweaking in our precommit account), but don't want to hold up this bug fix.

joshelser:
> Chatting with Ryan -- we're both baffled as to why test-deploy has started failing on a grant. The sql being run in deploy.py hasn't changed and this PR shouldn't affect anything to do with the app package. I'll look at it later (maybe something needs tweaking in our precommit account)

Fixed in #180. Just a misconfiguration on our testing account which I've fixed by hand.

PR #181, opened 2023-08-22
Overcome the 10k limit on the QUERY_HISTORY table function

joshelser:
bb73e11b07a4893a675f211270d442fe0684e158 is the only new change, the rest are duplicated by #176 

Was trying a couple of different approaches to try to make this work.

* I didn't think that a recursive CTE would actually work because I'd need something like LAST_VALUE and the documentation says that window functions are not allowed in the recursive clause https://docs.snowflake.com/en/user-guide/queries-cte#recursive-ctes-and-hierarchical-data
* I thought about writing a stored procedure in javascript but remembered that you can't run a query in a stored procedure with owner's rights that reads query_history without getting an error about "current_user()"  e.g. https://community.snowflake.com/s/article/Stored-Procedure-fails-with-error-Requested-information-on-the-current-user-is-not-accessible-in-stored-procedure

The next best thing I could think of was to enumerate each hour in the 24 hours of every day, and `UNION ALL` them together.

I had tried to generate the range of hours rather than enumerate them all, but I couldn't figure out how to get that to work passing it into the query_history table function, e.g.
```
with dates as (
    SELECT
        timestampadd('hour', rn, date_trunc('day', current_timestamp)) as start_range,
        timestampadd('hour', rn + 1, date_trunc('day', current_timestamp)) as end_range
    FROM (
            SELECT row_number() OVER (ORDER BY seq4()) - 1 AS rn
            FROM table(generator(rowcount => 24))
        ) AS g
)
 select
        qh.total_elapsed_time,
        qh.credits_used_cloud_services,
        qh.warehouse_size,
        qh.user_name,
        qh.role_name
    from dates d, table(information_schema.query_history(RESULT_LIMIT => 10000, END_TIME_RANGE_START => d.START_RANGE, END_TIME_RANGE_END => d.END_RANGE)) as qh;
```

maybe it's possible to make this work by wrapping the snowflake query_history table function into one that we make? (I feel like the arguments into information_schema.query_history is the reason this doesn't work) as https://docs.snowflake.com/en/sql-reference/functions-table is similar and says it should work.

Any thoughts?

### DIFF
--- a/bootstrap/007_probes.sql
+++ b/bootstrap/007_probes.sql
@@ -67,9 +67,10 @@ CREATE OR REPLACE PROCEDURE ADMIN.UPDATE_PROBE_MONITOR_RUNNING()
     EXECUTE AS OWNER
 AS
 BEGIN
-    let enable boolean := (select count(*) > 0 from internal.probes where cancel or notify_writer or length(notify_other) > 3);
+    let probes_enabled boolean := (select count(*) > 0 from internal.probes where cancel or notify_writer or length(notify_other) > 3);
     let configured boolean := (select count(*) > 0 from internal.config where key = 'post_setup');
-    if (enable and configured) then
+    let quota_enabled boolean := (select internal.is_consumption_enabled());
+    if ((probes_enabled OR quota_enabled) and configured) then
         execute immediate 'alter task tasks.PROBE_MONITORING resume';
     else
         execute immediate 'alter task tasks.PROBE_MONITORING suspend';
--- a/bootstrap/021_external_functions.sql
+++ b/bootstrap/021_external_functions.sql
@@ -97,6 +97,39 @@ EXCEPTION
         RAISE;
 END;
 
+-- Quota reporting to Sundeck EF
+BEGIN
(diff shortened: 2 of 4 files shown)


joshelser:
Also talked with Ryan about this. We have a strong aversion to doing some super-heavy or complicated tracking and are noodling over some ideas on how to simplify. We sketched out this idea:

* Store an aggregated credits-used in an internal table `credits_used` that enumerates credits used by each role and user by hour of the day (e.g. 0-23, no date component), with a row for each hour + user/role name + credits_used.
* Supplement the QUERY_HISTORY_MAINTENANCE task such that it computes the last hour's worth of query_history costs using account_usage.query_history (keeping in mind the 45min delay over account_usage.query_history), and update that hour's credit usage in the `credits_used` table
* On the cost control monitoring task, we use the information_schema.query_history table function to capture the last 15 to 75 minutes of missing data (covering the gaps from the last run of QUERY_HISTORY_MAINTENANCE and the delay on account_usage.query_history), merge that with contents of the `credits_used` table

joshelser:
overtaken by #190

PR #182, opened 2023-08-23
Ops center 110 migration

jinfengni:


bootstrap/090_post_setup.sql
```diff
@@ -68,6 +68,13 @@ call INTERNAL.POPULATE_PREDEFINED_LABELS();
 -- does not have user-created labels.
 call INTERNAL.INITIALIZE_LABELS();
 
+-- Migrate predefined labels to user's labels if user
+-- 1) does not make any change to predefined label,
+-- 2) and does not create new user label
+-- after last install/upgrade of APP
+-- parameter 300 is the timestamp difference when a predefined label is regarded as an old one.
```
rymurr, line 75: is this time time between sundeck updates? eg predefined labels older than 300 seconds in the predefined labels table are considered old?

If yes, I think making this longer is useful, The fastest we could do an update of the marketplace app is on the order of an hour

jinfengni: This parameter is to decide the predefined labels older than 300 seconds are considered old (imported from last install/upgrade of app). 

I changed from 300 seconds to 2 hours (7200 seconds).

Here are the logic for migration of predefined labels:
- predefined label has a `LABEL_CREATED_AT`, which captures the timestamp the row is inserted to `predefined_labels` table.  It will not change when we update the predefined label in new install / upgrade.
- `INTERNAL.INITIALIZE_LABELS` will import predefined labels into `labels` table, with `LABEL_MODIFIED_AT` set to be predefined lable's `LABEL_CREATED_AT`. 
-  The condition `predefined_label.LABEL_CREATED_AT = label.LABEL_MODIFIED_AT` is to indicate whether predefined label is sync with user's label.
- In upgrade (`INTERNAL.MIGRATE_PREDEFINED_LABELS`), only when the old `predefined_labels` are all sync with all the user's labels, we import the latest set of `predefined_labels` into `labels` table.  If user makes any change (delete or modify a predefined label, add a new user label),  migration will skip.

jinfengni:
@rymurr I rebased this PR after #172 is merged.  Please take another look when you get a chance. Thanks!

rymurr (approved):

PR #186, opened 2023-08-23
Add LABEL_MODIFIED_AT to predefined_labels table if it's created when…

jinfengni:
… labels did not have LABEL_MODIFIED_AT column


joshelser (approved):
Thanks for the quick fix, Jinfeng.

PR #190, opened 2023-08-24
Overcome the 10k row limit from information_schema.query_history for cost control

joshelser:
The last piece of #125 

At the end of the hourly QUERY_HISTORY_MAINTENANCE task, we will summarize the account_usage.query history for the last "full" hour (an "hour bucket"). Instead of tracking that hour by a complete timestamp, we only store the hour (0-23). Each row contains the sum of credits of one user or role in that hour bucket. Any previous rows in that same hour are deleted first before adding new rows in a transaction.

The cost monitoring task now only reads the "newest" query history data from information_schema.query_history, merges this credit usage summary with the data from the aggregated credit use table maintained by the query_history_maintenance task, and reports it to Sundeck.

This has a number of benefits:
* avoids the 10k rows limit from information_schema.query_history
* reading account_usage.query_history is much faster for accounts with large numbers of queries
* query_history_maintenance is already running hourly, so doing this little extra bit of work is super quick
* cost_control_monitoring becomes much quicker because it has significantly less work to do


bootstrap/041_quotas.sql
```diff
@@ -1,16 +1,16 @@
 
-create table if not exists internal.quota_task_history(start_time timestamp_ltz, end_time timestamp_ltz, credits_used object, outcome string);
+create table if not exists internal.quota_task_history(start_time timestamp, end_time timestamp, credits_used object, outcome string);
```
joshelser, line 2: This is a breaking change. We cannot do a normal migration we cannot `alter table` to change the datatype from `timestamp_ltz` to `timestamp`.

Currently, the only accounts which might have this table are sundeck accounts, so I think it best to not try to maintain some migration process.

rymurr: ive not thought too much about it but how are we handling timezones here?

bootstrap/041_quotas.sql
```diff
@@ -30,22 +30,116 @@ costed_queries as (
...
-    select 'users' as kind, object_agg(name, usage_map) as credits_used from user_usage union all select 'roles', object_agg(name, usage_map) from role_usage
 )
--- Aggregate into a final map {'users': {..}, 'roles': {..}}
-select object_agg(kind, credits_used) as daily_quota from usage_aggr;
+select name, persona, credits_used from usage;
 $$;
     return s;
 END;
+
+create table if not exists internal.aggregated_hourly_quota(hour_of_day integer, name string, persona string, credits_used float, last_updated timestamp);
```
rymurr, line 47: I am not sure I like this. why not just use the table function? then we don't have to mess around w/ edge cases. At the restriction that 10k rows in an hour max. Thats a good trade off to avoid the below code.

rymurr (commented):
Two main thoughts:
1. Is there a simpler algorithm
2. Do we need historical tracking

Lets chat when you are awake

joshelser:
> ive not thought too much about it but how are we handling timezones here?

The goal with timezones I had was to use only use timestamp_ntz (to which, timestamp is an alias to). There was one edge case I was chasing in that `current_timestamp()` returns a `timestamp_ltz`. I think I squashed that fully before publishing this PR. My expectation is that using timestamp_ntz should return timestamps that align to when the task runs that do not shift by timezone if you alter the user/session parameters for timezone.

bootstrap/041_quotas.sql
joshelser: This is the last piece I have to figure out, and I'm not entirely sure what is "correct" to do. So far, I've just been making sure that the timestamps are consistent (i.e. the hourly bucket matches when the task ran)

bootstrap/041_quotas.sql
```diff
@@ -1,51 +1,105 @@
...
+                   role_name,
+                   end_time
+            from table (information_schema.query_history(
+                   RESULT_LIMIT => 10000,
+                   -- TODO the function demands TIMESTAMP_LTZ. What does that actually do to our TIMESTAMP_NTZ?
+                   END_TIME_RANGE_START => TO_TIMESTAMP_LTZ(:qh_func_start_time),
+                   END_TIME_RANGE_END => current_timestamp()))
+        ), costed_queries as (
+            select
+                greatest(0, total_elapsed_time) * internal.warehouse_credits_per_milli(warehouse_size) +
```
jacques-n, line 46: I think this also needs to look at warehouse id and you should use the function that does both.

bootstrap/090_post_setup.sql
```diff
@@ -170,22 +170,49 @@ CREATE OR REPLACE TASK TASKS.COST_CONTROL_MONITORING
...
     sql STRING;
-    start_time timestamp_ltz default current_timestamp();
+    start_time timestamp default current_timestamp();
 BEGIN
     SYSTEM$LOG_DEBUG('starting task to monitor user and role cost');
-    -- Get and run the query to compute the daily quota usage by user and role.
-    call internal.get_daily_quota_select() into :sql;
+
+    -- Update internal.aggregated_hourly_quota with the latest usage data from the QH table function.
+    CALL INTERNAL.REFRESH_HOURLY_QUERY_USAGE_SQL(:start_time) into :sql;
```
jacques-n, line 178: i don't think it makes sense to use query generation here. none of the actual query executed needs to be dynamically generated. as best i can tell, a simple udtf would suffice for the core sql and then the merge could just be here as I think it is pretty simple. We should try to avoid dynamic sql unless we have to use it.

joshelser: > none of the actual query executed needs to be dynamically generated

You cannot query the information_schema.query_history in an owner's rights stored procedure because Snowflake fail. Frustrating :\

bootstrap/041_quotas.sql
joshelser: thanks! I had missed these functions. Makes sense.

bootstrap/090_post_setup.sql
jacques-n: Yes, understood. My comment stands. Let's avoid dynamic sql unless necessary. You're patterning after how probes work but probes must do dynamic sql. I don't see that same necessity here. You're right that we can't create a udtf except from a task context. I remember this from before. That being said, it still is best to avoid dynamic sql.

joshelser: Alright, thanks for the feedback. Let me try it out.

bootstrap/041_quotas.sql
jacques-n: I propose that we exclude serverless tasks from this calculation and document as such. 

Let's use the already defined "is_serverless_task" function to decide if it should be considered.

joshelser: > I propose that we exclude serverless tasks from this calculation and document as such.

SGTM

> Let's use the already defined "is_serverless_task" function to decide if it should be considered.

Looks like INFORMATION_SCHEMA.QUERY_HISTORY does not return the `WAREHOUSE_ID` column that ACCOUNT_USAGE.QUERY_HISTORY has, and that `INTERNAL.IS_SERVERLESS_WAREHOUSE(NUMBER)` relies on.

Cross-referencing some queries across the table function and the view, we check the prefix of the `WAREHOUSE_NAME` as another way to check this. I went looking for some other functions which might help determine this more reliably but didn't find anything.

I've introduced a new UDF with the same name that accepts a `WAREHOUSE_NAME` instead of the `WAREHOUSE_ID`

joshelser:
re: timezones and timestamps:

From what I can read, we don't want to store/use TIMESTAMP_LTZ as it would result in non-deterministic code that is based on the account's current TIMEZONE. Instead, we would want to use TIMESTAMP_TZ so that, if there are users across multiple timezones using the same OpsCenter app, they can reason about when the COST_CONTROL_MONITORING task ran (converting to their timezone, if desired). The challenge is that we have to use a TIMESTAMP_LTZ for INFORMATION_SCHEMA.QUERY_HISTORY and we want to avoid any "gaps" in computation due to timezone conversions.

I've updated this PR to:

When the COST_CONTROL_MONITORING task runs, it will get the `current_timestamp()` based on the account's current `TIMEZONE` parameter and then the task will qualify it with that timezone (making it a TIMESTAMP_TZ) .

Henceforth, we deal in terms of TIMEZONE_TZ to let users `CONVERT_TIMEZONE` from the account's default timezone to their timezone of choice. When we call `INFORMATION_SCHEMA.QUERY_HISTORY`, we cast the TIMESTAMP_TZ back to a TIMESTAMP_LTZ but this does not shift the absolute time of the TIMESTAMP.

I think this is easily explainable and avoids confusion about "when" we've read from query history.

bootstrap/090_post_setup.sql
```diff
@@ -169,33 +169,117 @@ CREATE OR REPLACE TASK TASKS.COST_CONTROL_MONITORING
...
     EXCEPTION
         WHEN other THEN
             SYSTEM$LOG_ERROR(OBJECT_CONSTRUCT('error', 'External function to report cost usage failed.', 'SQLCODE', :sqlcode, 'SQLERRM', :sqlerrm, 'SQLSTATE', :sqlstate));
-            quota_outcome := :sqlerrm;
+            quota_outcome := :quota_outcome || '. Failed to report quota usage to Sundeck ' || :sqlerrm;
     END;
 
     -- Log the results locally
-    insert into internal.quota_task_history select :start_time, current_timestamp(), :quota_usage, :quota_outcome;
+    INSERT INTO internal.quota_task_history SELECT :start_time, current_timestamp(), :quota_usage, :quota_outcome;
```
joshelser, line 276: for some reason, `:quota_outcome` is always `null` now and I can't yet figure out why.

joshelser: `varchar || null` nets a `null`. Whoops.

rymurr:
I am not sure I understand the timezone stuff...everything else looks good.

joshelser:
> If everything is in LTZ then its not clear to me what the problem is.

The blog states it better than I can -- the surprise in working with LTZ is that it's easy to introduce bugs because of that dynamic timezone calculation.

I don't think there is an immediate problem in TZ vs. LTZ, just the potential for later confusion/bugs. It sounds like you think returning LTZ is less complicated and better? I think just need to make the call (the important piece was that we weren't "skipping" any ranges of QH data being read, IMO) -- I could see TZ avoiding errors in the future, but I also don't think that is imminent.

rymurr:
For me playing with timezones complicates things now and is at least more error prone. I can't envision any bugs that will occur from using ltz in the future (over tz) but I can definitely think of bugs related to making the tz change by hand in this code. I, of course, could be missing something obvious.

joshelser:
> I can't envision any bugs that will occur from using ltz in the future (over tz) but I can definitely think of bugs related to making the tz change by hand in this code. I, of course, could be missing something obvious.

It may have been a repost of the same blog post, but the majority of my hesitation came from this blog post which warned "[LTZ will make your stored procedure non-deterministic]". I'll switch this back to LTZ types.

rymurr (approved):

rymurr (approved):

PR #192, opened 2023-08-25
Fix topn ordering

rymurr:
the `topn` function was behaving weird wiiht `ngroup` this ensures that the top n are always the largest in a given dimension


joshelser (approved):
Indeed, looks much better with your patch.

joshelser (approved):

PR #197, opened 2023-08-28
Ensure we dont read tables before they are settled

### LINKED ISSUE #177: Warehouse Activity reports shows costs 100x over Query Activity report on one day
In one of our internal accounts, I see that the reported cost on a given day (Aug 10th) by warehouse is ~100x over the equivalent days' cost on the query activity report.

I think @rymurr might have been doing some work on snowpark that day, as the majority of the cost is associated with a similarly named warehouse.

I'm guessing it's accurate that there were no "traditional" queries being run at the same time, however, it's a little scary that I can't find an association of where this cost is coming from through opscenter. Maybe we could make a report which can describe snowpark usage?

rymurr:
closes #177

### DIFF
--- a/bootstrap/090_post_setup.sql
+++ b/bootstrap/090_post_setup.sql
@@ -14,12 +14,15 @@ BEGIN
                 from information_schema.columns
                 where table_catalog = current_database() and table_schema in ('ACCOUNT_USAGE', 'ORGANIZATION_USAGE')
                 group by table_name, table_schema
-              )
+              ), delays as (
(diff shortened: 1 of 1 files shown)


rymurr:
unfortunately, I think to get an accurate calculation after this has been merged will require resetting the internal data set. Any data past this fix will work.

joshelser:
> This appears to be some missing data in the account in question, starting on Aug 8 the data volume in enriched query history doesn't match up with account usage.

Extrapolating from your last comment on the issue -- we happened to see some data in one of the account_usage views before the correlating data in a different account_usage view, which prevented us from properly assigning the costs?

rymurr:
not quite. The data in `snowflake.account_usgae.query_history` is possibly delayed up to 45 min. The problerm is that rows that get added can be added below max (end_time) and we keep max(end_time) as a high water mark

joshelser:
> The problerm is that rows that get added can be added below max (end_time) and we keep max(end_time) as a high water mark

Thanks for explaining, I think I understand now.

joshelser (approved):

joshelser (approved):

PR #198, opened 2023-08-28
Pin streamlit to 1.22.x, should fix cypress

joshelser:
This matches what is currently in snowflake.information_schema.packages

When I talked to Vicky on Friday, she had figured out that the version of streamlit in SF had changed since end of June when we last checked. Cypress was failing because of a streamlit stacktrace being printed, which I believe stems from the wrong dep version string for streamlit in pyproject.toml (we wanted "nothing more recent than 1.16" but had "anything newer than 1.16".

The rest of us hadn't noticed because of the poetry.lock we already had (likely at streamlit 1.22.x or 1.24.x)

### DIFF
--- a/app/ui/environment.yml
+++ b/app/ui/environment.yml
@@ -6,3 +6,4 @@ dependencies:
 - plotly
 - altair
 - snowflake-native-apps-permission
+- streamlit=1.22.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 [tool.poetry.dependencies]
 python = "~3.8"
 plotly = "~5.11"
-streamlit = "^1.16.0"
+streamlit = "~1.22"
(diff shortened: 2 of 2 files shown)


rymurr:
I am using streamlit `1.16` and last I checked that was the only package available for python `3.8`. Its not clear to me why this simple fix works?

vqmarkman (dismissed):
@joshelser , I made the same fix locally on Friday and nothing failed for me. Did not occur to me to try it in automation. Thanks for the fix.

Some more info: ran poetry upgrade and my poetry.lock is showing and tests don't fail:

```
[[package]]
name = "streamlit"
version = "1.24.0"
```

rymurr:
can we pin these values in `app/ui/environment.yml`? Pinning in poetry isn't going to help a whole lot if we aren't pinning in what is deployed to snowflake.

rymurr (approved):

PR #202, opened 2023-08-28
Opscenter-201: Exclude streamlit queries in predefined probe condition

jinfengni:


bootstrap/007_probes.sql
```diff
@@ -291,8 +291,8 @@ BEGIN
     USING (
         SELECT *
         from (values
-                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp())'),
-                ('Big Readers', 'bytes_scanned > 10000000000')
+                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp()) AND NOT (query_text ilike \'execute streamlit%\')'),
+                ('Big Readers', 'bytes_scanned > 10000000000 AND NOT (query_text ilike \'execute streamlit%\')')
```
joshelser, line 295: Do native apps have a large number of bytes scanned?

jinfengni: for now, probably not. I just want to avoid the case similar to "long queries", where someday the app uses large number of scan.

bootstrap/007_probes.sql
```diff
@@ -291,8 +291,8 @@ BEGIN
     USING (
         SELECT *
         from (values
-                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp())'),
-                ('Big Readers', 'bytes_scanned > 10000000000')
+                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp()) AND NOT (query_text ilike \'execute streamlit%\')'),
```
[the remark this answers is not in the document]
jinfengni: let me check the queries submitted through the app, and see if there is a better way.

bootstrap/007_probes.sql
joshelser: Yeah, I just double-checked and it looks like `bytes_scanned` is always zero. I don't think this is super distracting from how this probe works, but wanted to be sensitive to adding too many "this but not that" kinds of probes.

bootstrap/007_probes.sql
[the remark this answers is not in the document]
joshelser: `QUERY_TYPE = 'EXECUTE_STREAMLIT'`?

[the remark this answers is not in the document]
jinfengni: yes, that's a better way.

bootstrap/007_probes.sql
jinfengni: makes sense. I will remove this pre-cautious condition for "Big Readers" probe.

jinfengni:
@rymurr @joshelser I made changed from the feedback.  Can you take another look if you get a chance? Thanks!

joshelser (approved):

joshelser:
Thanks Jinfeng!

PR #205, opened 2023-08-29
Opscenter-135: qlike support optional matching parameters

jinfengni:


jinfengni:
Testing of `tools.qlike()` function requires manually configuration of sundeck token.  Run the following query to verify the new functions.

1.  without parameter
```
select tools.qlike('select * from (select * from a.b.c)', 'select * from a.b.c');
+---------------------------------------------------------------------------+
| TOOLS.QLIKE('SELECT * FROM (SELECT * FROM A.B.C)', 'SELECT * FROM A.B.C') |
|---------------------------------------------------------------------------|
| True                                                                      |
+---------------------------------------------------------------------------+
```

2. Using 'p' for CONTAINS matching ( which is the default behavior)
```
select tools.qlike('select * from (select * from a.b.c)', 'select * from a.b.c', 'p');
+--------------------------------------------------------------------------------+
| TOOLS.QLIKE('SELECT * FROM (SELECT * FROM A.B.C)', 'SELECT * FROM A.B.C', 'P') |
|--------------------------------------------------------------------------------|
| True                                                                           |
+--------------------------------------------------------------------------------+
```

3. Using `a` for COVERS matching
```
select tools.qlike('select * from (select * from a.b.c)', 'select * from a.b.c', 'a');
+--------------------------------------------------------------------------------+
| TOOLS.QLIKE('SELECT * FROM (SELECT * FROM A.B.C)', 'SELECT * FROM A.B.C', 'A') |
|--------------------------------------------------------------------------------|
| False                                                                          |
+--------------------------------------------------------------------------------+
```

Before this PR's change,  using a matching parameter in QLIKE() function would hit an error.
```
select tools.qlike('select * from (select * from a.b.c)', 'select * from a.b.c', 'a');
001044 (42P13): SQL compilation error: error line 1 at position 7
Invalid argument types for function 'QLIKE': (VARCHAR(35), VARCHAR(19), VARCHAR(1))
```

joshelser:
> Testing of tools.qlike() function requires manually configuration of sundeck token

Just double-checking, you mean that it requires the Sundeck account to be linked to OpsCenter? (either the Snowflake SSO or copying in the Sundeck API key are sufficient)

jinfengni:
> it requires the Sundeck account to be linked to OpsCenter

Yes, the external function will go through a sundeck tenant (see [token used here](https://github.com/sundeck-io/OpsCenter/blob/main/bootstrap/021_external_functions.sql#L221-L225)).  The link between the two is communicated using a token. 

I was thinking to add a unit test similar to `test_labels_procs.py`. But realized I do not have a way to setup token in the tests.

joshelser:
> I was thinking to add a unit test similar to test_labels_procs.py. But realized I do not have a way to setup token in the tests.

Yeah, it's not possible to create a unit test to cover this since it requires the external function to Sundeck. I don't think we have OpsCenter set up in our integration testing accounts, but we could do that. Let's talk tmrw if you have some time :)

joshelser (approved):
Nice. I thought there was going to be a migration problem, but then realized that this does not require a migration fix! That's wonderful.

It's probably worth putting together a short documentation page under https://docs.sundeck.io/opscenter/ to tell users that QLIKE is available in OpsCenter. I see we don't have anything now.

rymurr:
> Nice. I thought there was going to be a migration problem, but then realized that this does not require a migration fix! That's wonderful.
> 
> It's probably worth putting together a short documentation page under https://docs.sundeck.io/opscenter/ to tell users that QLIKE is available in OpsCenter. I see we don't have anything now.

There is some docs already: https://docs.sundeck.io/opscenter/example-labels/

joshelser:
> There is some docs already: https://docs.sundeck.io/opscenter/example-labels/, given the complexity around qlike I am not inclined to publicize it more

Gotcha, I missed it in the examples labels page. I have been finding useful stuff in the TOOLS schema that I didn't know was there. That's as far as my thinking had gone ^.^

joshelser:
> did i miss any in the docs. Its my thinking as well and I tried to document it all

The `TOOLS.MODEL_*` functions are the ones that are obtuse to me (not to derail Jinfeng's pr)

rymurr:
> The TOOLS.MODEL_* functions are the ones that are obtuse to me (not to derail Jinfeng's pr)

yeah, fair point. They are used in the dbt report if that helps. I had not thought of using them in labels....interesting

jinfengni:
Thanks for the link to the doc. I'll add one example of the qlike with matching parameter in the doc.

jinfengni:
Opened a new ticket while trying to use `tools.qlike()` in label condition: https://github.com/sundeck-io/OpsCenter/issues/211 .  I'll take a look to see the cause of it.

PR #208, opened 2023-08-30
Add button in settings to reload predefined p&l

### LINKED ISSUE #204: Add default probes & labels upon request
If I'm one of a handful of users who installed (and didn't uninstall) OpsCenter before we provided default labels and probes, it'd be great if I would click a button and get them added. This isn't "high" priority because we don't have many users who installed. However, when we reach out to those existing users, it'd be nice if we could help them with some sample labels and probes. 

I'm curious how difficult this would be. thoughts?

rymurr:
We add a button to settings page which will reload all predefined probes and labels from the static table

closes #204


app/ui/pages/10_Settings.py
```diff
@@ -276,3 +283,16 @@ def expander(num: int, title: str) -> st.expander:
         connection.execute("call internal.refresh_queries(true);")
         bar.progress(100, text="All events refreshed.")
         msg.info("Reset Complete.")
+
+with initial_probes:
+    st.title("Pre-Configured Probes and Labels")
+    do_reset = st.button("Re-load all preconfigured labels and probes.")
```
joshelser, line 289:
```suggestion
    st.title("Pre-Configured Probes and Labels")
    st.text("OpsCenter includes a number of helpful labels and probes that are created when you have none defined. OpsCenter will not overwrite any existing labels and probes you have defined. Click the button below to reset all labels and probes to the defaults provided by OpsCenter.")
    do_reset = st.button("Re-load all preconfigured labels and probes.")
```

Something like this to warn that this is destructive and explain why you might want to click this button?

joshelser (approved):
thank you!

PR #210, opened 2023-08-30
Add initial probes which detect high credit usage

rymurr:


bootstrap/018_warehouse_updates.sql
```diff
@@ -63,5 +64,29 @@ BEGIN
...
+    END;
+    SYSTEM$LOG_INFO('Finished warehouse size table refresh');
     CALL INTERNAL.SET_CONFIG('WAREHOUSE_EVENTS_MAINTENANCE', CURRENT_TIMESTAMP()::string);
 END;
+
+CREATE OR REPLACE FUNCTION TOOLS.APPROX_CREDITS_USED(warehouse_name varchar, start_time timestamp, end_time timestamp)
+RETURNS NUMBER
+AS
+$$
+    TOOLS.WAREHOUSE_CREDITS_PER_MILLI((select warehouse_size from internal.warehouse_size_mapping t where t.warehouse_name = warehouse_name))*timestampdiff(millisecond, start_time, end_time)
```
joshelser, line 84: I see that `CREDITS_USED_CLOUD_SERVICES` is included in the query_history table function. Not sure how difficult it is to use a bunch of cloud services credits compared to warehouse credits. Intentional to omit `CREDITS_USED_CLOUD_SERVICES` here?

bootstrap/007_probes.sql
```diff
@@ -291,17 +291,19 @@ BEGIN
     USING (
         SELECT *
         from (values
-                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp()) AND NOT QUERY_TYPE = \'EXECUTE_STREAMLIT\''),
-                ('Big Readers', 'bytes_scanned > 10000000000')
-             )) s (name, condition)
+                ('Long Queries', 'start_time < dateadd(minute, -10, current_timestamp()) AND NOT QUERY_TYPE = \'EXECUTE_STREAMLIT\'', False),
+                ('Big Readers', 'bytes_scanned > 10000000000', False),
+                ('Costs 10 Credits', 'tools.approx_credits_used(warehouse_name, start_time) > 10', True),
+                ('Costs 50 Credits', 'tools.approx_credits_used(warehouse_name, start_time) > 50', True)
```
joshelser, line 297: I think we would want these to be greater-than-or-equal-to instead of greater-than. `approx_credits_used` is going to return a NUMBER (integer), which would mean the queries we match are actually those which cost at least 11 or 51 credits

rymurr: NUMBER is ~DECIMAL so it will have a decimal.

joshelser: ah, I had just done `SELECT 10.0::NUMBER` and it truncated it. My sanity check was just invalid :)

joshelser (approved):

PR #206, opened 2023-08-29
Correct the ordering of DAILY and COMPLETE for PERIOD_PLUS

joshelser:
This bug causes the binning of queries by day to be backwards; the "total" query row was sent to the DAILY view and the rows per day for a query were sent to the COMPLETE view.

Verified on existing installations that this fix only applies to newly captured queries and does not rewrite old queries (confirmed by the metrics in internal.task_query_history table)

Closes #143

Some "proof" for the verification around not-regenerating the entire query history. here is the output from `select * from internal.task_query_history`

```
RUN	SUCCESS	INPUT	OUTPUT
2023-08-29 15:17:33.511	TRUE		{   "attempted_migrate": true,   "new_INCOMPLETE": 2,   "new_closed": 21124162,   "new_records": 21124164,   "newest_completed": "2023-08-29 14:32:33.459",   "oldest_running": "2023-08-29 14:32:33.459" }
2023-08-29 15:33:53.493	TRUE	{   "attempted_migrate": true,   "new_INCOMPLETE": 2,   "new_closed": 21124162,   "new_records": 21124164,   "newest_completed": "2023-08-29 14:32:33.459",   "oldest_running": "2023-08-29 14:32:33.459" }	{   "attempted_migrate": true,   "new_INCOMPLETE": 2,   "new_closed": 1572,   "new_records": 1574,   "newest_completed": "2023-08-29 14:48:41.623",   "oldest_running": "2023-08-29 14:48:41.623" }
2023-08-29 15:41:01.515	TRUE	{   "attempted_migrate": true,   "new_INCOMPLETE": 2,   "new_closed": 1572,   "new_records": 1574,   "newest_completed": "2023-08-29 14:48:41.623",   "oldest_running": "2023-08-29 14:48:41.623" }	{   "attempted_migrate": true,   "new_INCOMPLETE": 2,   "new_closed": 884,   "new_records": 886,   "newest_completed": "2023-08-29 14:55:42.678",   "oldest_running": "2023-08-29 14:55:42.678" }
```

Row 1: unpatched fresh computation after a "Reset" (v1.17)
Row 2: first run after fixing this bug for QH and WH
Row 3: a second run to verify that subsequent runs also don't "rewind"

The `OUTPUT` column on rows 2 and 3 are demonstrating that we are not re-reading all ~21M rows in this account's query_history view.

Issue #103, opened 2023-07-31
Refresh data post table migration


rymurr:
After table schemas are changed (eg #101 ) we will have new columns in the reporting tables but the data from the backing tables (#88) or newly added calculated tables (#86) will be missing. Only newly arrived data will get the new fields. In this case we should add a banner to the main page which informs the user of new features being available and that they should go to the settings page and reset their data in order to pick up the new columns.

It is important that the banner is contextual and useful. Eg **NOT** "Go to settings -> reset" to get new data". Rather "In our latest release we added x,y,z (maybe a link). By resetting your data you will get these new features which will allow you to a,b,c (maybe link)"

rymurr:
thinking it might be a good idea ot create an internal table with `VERSION, DESCRIPTION, AFFECTED_PAGES,REFRESH_TIMESTAMP`
We can write into this on release (with refresh_timestamp = `NULL`) and read from it when we showw a page. If hte page is in the table then we show the banner. When refresh is ticked insettings we update all null timestamps.

joshelser:
Yeah, that's about what I was thinking. Thanks for writing it out!

PR #228, opened 2023-09-01
Correct the ordering of DAILY and COMPLETE for PERIOD_PLUS

### LINKED ISSUE #143: Possible wrong direction on condition in QUERY_HISTORY_COMPLETE_AND_DAILY
line 10 has `IFF(index = PEROID_PLUS...` it is possible that this condition has been reveresed and is causing dups in #31

joshelser:
Re-application of #206 with extra logic to "fix" the old data through the user-facing views.

For QH and WEH tasks, we do the following:
* Record the END_TIME/SESSION_END for the next task run in `internal.config` (all rows materialized _after_ this time are correct and we don't want to invert the RECORD_TYPE on these)
* Update the views (daily, complete) to:
* * exclude any rows which span more than one day in the existing `select` (are subject to this bug)
* * union a new `select` which includes the rows that span more than one day, inverting `RECORD_TYPE`

Closes #143


joshelser:
I talked about this PR with @rymurr this morning among other things.

We both felt like this current PR, while effective at "fixing" the materialized data, is not a sustainable pattern long-term. We are lucky with this change that we _can_ identify and fix the data -- Ryan's fix in #197 was not so lucky.

For fixes to materialized data that require a full re-materialization, #103 (assigned to me for this sprint) will inform users of the benefit of clicking the "reset" button, and enumerate the positives of doing so.

The outstanding question is how do we address this PR?
* Ryan believed that adding custom logic to fix the old data was not worth the debt. This would not fix materialized data (I'd remove that part of this change and rewind back to an earlier change that just fixes newly materialized data).
* Alternatively, we could treat this like a schema migration and add a procedure to re-write the offending query_id's in the materialization. This would still be one-off.

I'll give @jacques-n some time to chime in, but I'm currently leaning towards Ryan's suggestion of deferring this to #103

jacques-n:
Isn't [this change](https://github.com/jacques-n/OpsCenter/commit/7851f64bcfa6abdae0f577dfc23b5fe3f8e5796b)

A much simpler way to fix this that also avoids a lot of debt or is there something I'm missing?

joshelser:
> a simpler way to fix this that also avoids a lot of debt or is there something I'm missing?

Yeah, that is true. I had been trying to avoid swapping `record_type` for single-day-spanning queries and got tunnel-vision around that. Let me try out your suggestion. I think it will work and be a little more concise (at the low cost of different 'enum' values for RECORD_TYPE).

Thanks!

joshelser:
Jacques' suggestion came out much cleaner. The only issues was having to alter the `RECORD_TYPE` column on the materialized tables from `VARCHAR(8)` to `VARCHAR`. Without this, the new values for `RECORD_TYPE` were rejected because they were longer than 8 chars.

rymurr (approved):
Much cleaner! Love it!

Issue #225, opened 2023-09-01
Probes test failure: index out of range


vqmarkman:
Happened in PR: https://github.com/sundeck-io/OpsCenter/actions/runs/6047474866/job/16413995185
During initial  loading of the page.

https://github.com/sundeck-io/OpsCenter/assets/104043363/c038005c-eef8-4414-be19-e28376c2d752

cc @rymurr

vqmarkman:
My conclusion is that this is OpsCenter bug and I can't reproduce it locally :(
The worst part about it is that video does not show last (most important part) of the stack trace.

@doronrosenberg , @susannahmarcus - is there any other way to debug this type of failure? Where should I be looking for the logs?

susannahmarcus:
I can't say much since I haven't worked on this much, but I can weigh in on the viewport. For future instances like this, we could make it so that it isn't cut off. Additionally, we might be able to add a check for the type of an error message and then have it spit that into the cypress log, so we can get more information. 

If you'd like to work on that together, feel free to reach out and we can try some stuff out locally 😊 👍

joshelser:
> The worst part about it is that video does not show last (most important part) of the stack trace.

Well, I think we know about where it is failing. It's inside the `_call(..)` method, so we can assume it's stuck on the snippet in probes.py
```
_ = self.snowflake.call(
            "INTERNAL.REPORT_ACTION",
            "probes",
            "list",
        )
```

`self.snowflake` here is a snowflake-snowpark-python `Session` instance, which, I'm guessing we're on the line https://github.com/snowflakedb/snowpark-python/blob/v1.7.0/src/snowflake/snowpark/session.py#L2447 inside of the Snowpark library.

I'd guess that we didn't get any rows back (for whatever reason) like #35 which caused this. Maybe we need to try to figure out why this `Session` seems to be returning bogus results sometimes?

vqmarkman:
@joshelser, the "interesting" part about this thing is ... of course this particular case (probes test) is failing in automation and is not failing locally for me. I'm not sure what the next steps here are ...

joshelser:
I wonder if this is related to https://github.com/sundeck-io/OpsCenter/commit/a2f5b67c61679daa43f0c3e5151e13f6112e5f89

[...]

PR #229, opened 2023-09-01
Issue #225 - bandaid to workaround failing cypress tests due to issue…

vqmarkman:
… described in issue #225

             Bit heavy handed, but I don't want this issue to block any further development.
             Will have a better fix when I find the solution.


joshelser (approved):
let's try this to get pre-commit happier

PR #232, opened 2023-09-03
Working through flaky tests.

vqmarkman:
Fixes single test that was failing for me locally all the time.


joshelser:
Which test was failing? Do you know why this change will stabilize that test or is this a guess that it will help?

I have no reason to think your change is wrong, but I'm also not sure if it's likely to help.

vqmarkman:
This is cypress specific fix, fixes labels test: "Menu: Labels. Create/Delete grouped labels".
I saw it fail in automation and I was able to reproduce it locally.

joshelser (approved):

Issue #169, opened 2023-08-11
Wire python tests to run in github actions


vqmarkman:

joshelser:
Hrm, missing pytest during the job.. somehow

joshelser:
```
FAILED unit/test_probe_procs.py::test_smoke_create_drop_probe - snowflake.connector.errors.ProgrammingError: 002003 (02000): Uncaught exception of type 'STATEMENT_ERROR' on line 26 at position 4 : Uncaught exception of type 'STATEMENT_ERROR' on line 7 at position 8 : SQL compilation error:
Task 'UNIT_TESTS_VERIFY_UNIT_TESTS_6112029882.TASKS.PROBE_MONITORING' does not exist or not authorized.
FAILED unit/test_probe_procs.py::test_create_probe_with_existing_name - snowflake.connector.errors.ProgrammingError: 002003 (02000): Uncaught exception of type 'STATEMENT_ERROR' on line 26 at position 4 : Uncaught exception of type 'STATEMENT_ERROR' on line 7 at position 8 : SQL compilation error:
Task 'UNIT_TESTS_VERIFY_UNIT_TESTS_6112029882.TASKS.PROBE_MONITORING' does not exist or not authorized.
FAILED unit/test_probe_procs.py::test_smoke_update_probe - snowflake.connector.errors.ProgrammingError: 002003 (02000): Uncaught exception of type 'STATEMENT_ERROR' on line 26 at position 4 : Uncaught exception of type 'STATEMENT_ERROR' on line 7 at position 8 : SQL compilation error:
Task 'UNIT_TESTS_VERIFY_UNIT_TESTS_61120[298](https://github.com/sundeck-io/OpsCenter/actions/runs/6112029882/job/16588600367?pr=257#step:9:299)82.TASKS.PROBE_MONITORING' does not exist or not authorized.
FAILED unit/test_probe_procs.py::test_create_probe_with_empty_string_name - snowflake.connector.errors.ProgrammingError: 002003 (02000): Uncaught exception of type 'STATEMENT_ERROR' on line 26 at position 4 : Uncaught exception of type 'STATEMENT_ERROR' on line 7 at position 8 : SQL compilation error:
Task 'UNIT_TESTS_VERIFY_UNIT_TESTS_6112029882.TASKS.PROBE_MONITORING' does not exist or not authorized.
FAILED unit/test_probe_procs.py::test_new_negative - assert False
```

[...]

Issue #36, opened 2023-07-12
Switch to Conda


joshelser:
We switched the project to using https://python-poetry.org/ in an attempt to be more "modern" in managing a Python project.

Since then, we found https://repo.anaconda.com/pkgs/snowflake/ which we think might have the actual versions of third party libraries that will be activated when the Native App runs inside of Snowflake. We've had issues due to getting a different version of plotly, and we want to avoid getting into more issues where the app runs correctly with local streamlit but with error when inside of Snowflake.

rymurr:
conda is how Snowflake recommend setting things up but it appears the anaconda channel is out of date compared to whats in Snowflake. eg streamlit 1.22 is available in snowpark but only 1.16 on the conda channel. However, I think conda would still be safer than using poetry

PR #233, opened 2023-09-04
Basic wiring of pytests into workflow

joshelser:
Uses conda and its own database via `devdeploy.py -s` (skipping materialization to keep costs and execution time down).

Rel #169

test/README.md was a little sparse on how to actually run the tests (e.g. what are the python dependencies, what needs to exist in snowflake), so I made some educated guesses to fill in the gap. If those are correct, I will update the README.

This passed for me in my fork https://github.com/joshelser/OpsCenter/actions/runs/6067975974/job/16460330862?pr=3

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -15,6 +15,44 @@ jobs:
           python-version: '3.8'
           cache: 'pip'
       - uses: pre-commit/action@v3.0.0
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          submodules: recursive
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.8'
+          cache: 'pip'
+      - name: install dependencies
+        run: |
+          pip install poetry
+          poetry install --with=dev
+      - name: Make unique names for this job
+        id: shell-command
+        run: |
+          SUFFIX=$(echo ${GITHUB_HEAD_REF} | sed "s/\./_/g" | sed "s/-/_/g" | sed "s\/\_\g")_${GITHUB_RUN_ID}
+          echo "OPSCENTER_DATABASE=UNIT_TESTS_$SUFFIX" >> $GITHUB_ENV
+      - name: Make snowsql config
+        env:
+          # Also uses OPSCENTER_DATABASE from above
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
+          SNOWFLAKE_USERNAME: ${{ secrets.SNOWFLAKE_USERNAME }}
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+        run: |
+          mkdir $HOME/.snowsql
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -15,6 +15,41 @@ jobs:
...
+        shell: bash -el {0}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          submodules: recursive
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: opscenter-test
+          environment-file: test/environment.yml
```
rymurr, line 32: why not use app/ui/environment.yml?

rymurr: and a `conda install pytest`

joshelser: Well, I guess two reasons

1. `streamlit` 1.16.0 is the latest available version in snowflake's anaconda channel (can probably work around that without too much pain)
2. `snowflake-native-apps-permission` also appears to be missing from snowflake's channel which makes no sense to me

I'm not sure why the permissions SDK package wouldn't be on the public-facing anaconda repo.

joshelser: Yeah, i'm not sure here. I can't find any mention of the permissions SDK on the public internet. The mismatch of information_schema.packages and the snowflake conda repo (1.22.0 the latest on the former, 1.16.0 on the latter) is another annoying wrinkle.

I'm a little worried about changing the "real" environment.yml and accidentally breaking OpsCenter releases. Since the set of dependencies is effectively disjoint (what we think is being used at runtime vs. what is found on snowflake's conda channel), what do you think of having separate environment.yml files for "product" and "test"? Maybe I've missed some other way forward?

I can also try reaching out to our Snowflake contacts and see if they can shed any light on this.

rymurr: The environment file is separate from the channel being used isn't it? Why do we have to use the public snowflake anaconda channel to use conda? Maybe I missed something?

We already have a poetry definition of the all and it's dependencies and a conda one. I don't really want to add a 3rd to maintain.

joshelser: We don't have to use Snowflake's conda channel, but I thought that was a goal behind switching to Conda (because we assume their Conda channel better matches what we'd get at runtime running inside Snowflake). Agree that if we don't care about using the snowflake channel, this becomes more straightforward.

Pyproject.toml file does not list the native app permissions sdk as a dependency which is why we currently don't have this problem.

I can appreciate not wanting more files declaring conflicting lists of dependencies. However, I'm not sure how we could use one environment.yml file for both inside and outside of Snowflake. I guess we could run some bash commands in the GitHub workflow to delete the permissions dependency line before we run unit tests? I couldn't find anything like an optional dependency feature with Conda which I expected to be able to do.

rymurr: For this change I think we have two options. Encode this into the pyproject.toml or figure out a better solution for conda. I am inclined to lean towards pyproject.toml until we have a better understandign of how snowflake is going to do its anaconda channel.

I def don't want to mix anaconda and poetry for our internal tests though...

joshelser: Sounds good. I'll back out the conda changes, use poetry for this PR, and we'll take on the conda migration in #36

Issue #213, opened 2023-08-30
Disable task and checkbox for "user limits" if Sundeck isn't connected


ron-sundeck:
Feature is "user limits"

Feature introduced a checkbox in settings to enable/disable a task that executes every 5 minutes. If opsCenter isn't connected to a Sundeck account, then the task shouldn't run and the checkbox to enable the task should be disabled.

PR #237, opened 2023-09-04
Conditionally start user limits maintenance after Sundeck link

joshelser:
* Rename everything from cost control to user limits
* Adds `internal.has_config(string) returns boolean` procedure
* Task is still created during finalize_setup, but is resumed only after one of the two sundeck "linking" tasks are complete.

ref #213


bootstrap/000_setup_base_tools.sql
```diff
@@ -39,6 +39,13 @@ BEGIN
     return config_value;
 END;
 
+CREATE OR REPLACE PROCEDURE internal.has_config(key string) RETURNS BOOLEAN LANGUAGE SQL
+    AS
+BEGIN
+    let config_value string := (SELECT value FROM internal.config WHERE key = :key);
```
rymurr, line 45: can you return `internal.get_config(:key) is not null`? Just a bit less code dup

joshelser: I can definitely `call internal.get_config(:key)` to avoid the `select` query, but it doesn't seem like I can compress it down to a one-liner (have to call the proc into a var and then check the var)

rymurr (approved):

PR #242, opened 2023-09-06
dbt report should be using labeled query history

rymurr:


joshelser (approved):
Does this change have impact? It looks like labeled_query_history is just a view created from enriched_query_history.

rymurr:
If a user selected a label filter the query would Jane bombed out.

PR #255, opened 2023-09-06
Revert "We should take column names from snowflake tables now that ou…

joshelser:
…r migration code works (#243)"

This reverts commit 0d88aae747c6f449274fa7143e566aaa5195782a.


vqmarkman (approved):
Let's revert for now to unblock others.

PR #258, opened 2023-09-07
Keep the unit-test job from failing the workflow

joshelser:
add some debugging. don't know why it's installing the other dev dependencies but not the pytest dependency. Maybe something weird in the caching on the github worker?

rel #169

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -17,6 +17,8 @@ jobs:
       - uses: pre-commit/action@v3.0.0
   unit-tests:
     runs-on: ubuntu-latest
+    # Temporarily allow this job to fail and not fail the whole workflow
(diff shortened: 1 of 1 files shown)


rymurr (approved):

joshelser:
Of course, the unit tests installed pytest this time.

My hunch is that it has somethign to do with the global (not inside the checkout) virtualenv, and pollution across the jobs.

.github/workflows/pull-request.yaml
```diff
@@ -30,6 +32,7 @@ jobs:
         run: |
           pip install poetry
           poetry install --with=dev
```
rymurr, line 34: technically the `--with dev` is not required...is that the issue?

joshelser:
Merged to unblock other PRs.

PR #260, opened 2023-09-07
Materialize once in pull-request workflow, then run all tests

joshelser:
…onitoring

We could also solve this by updating `ADMIN.UPDATE_PROBE_RUNNING()` to conditionally check for the probe_monitoring task to exist. My gut reaction is that adding these kind of checks may turn into a game of whack-a-mole. Admittedly, I'm not sure what actual change in cost we will see as a result of this additional logic.

Maybe I use opscenter to figure that out? :P 

rel #169


.github/workflows/pull-request.yaml
```diff
@@ -50,7 +50,8 @@ jobs:
           mkdir $HOME/.snowsql
           envsubst < deploy/opscenter.config > $HOME/.snowsql/config
       - name: Run deploy/devdeploy.py (without triggering materializations)
-        run: poetry run python deploy/devdeploy.py --profile opscenter --skip-finish-setup
+        # Some unit tests call procedures which expect admin.finalize_setup() to have been run.
```
rymurr, line 53: doesn't this materialize all the data too? Expensive isn't int?

rymurr: not sure an alternative, just run `finalize_setup` w/o all the other stuff?

rymurr: can we run this in the cypress job so we re-use the db?

joshelser: > not sure an alternative, just run finalize_setup w/o all the other stuff?

Not an alternative, I think, as the tasks would be manually triggered at the end of `finalize_setup()`

> can we run this in the cypress job so we re-use the db?

That makes sense to me. We can do one, real deploy (with data), and run many tests against it.

joshelser:
Shows up something like https://github.com/joshelser/OpsCenter/actions/runs/6116662251?pr=4

there's a latent pytest failure even after all of this (I assume it's just busted and no one was watching)
```
unit/test_probe_upgrade.py:64: AssertionError
--------------------------- Captured stdout teardown ---------------------------

[INFO] Teardown after running tests in the session
[INFO] SQL stmt to find all the labels: select name from INTERNAL.LABELS where name like '%2023-09-08-01-46-00%'
[INFO] SQL in delete function: select name from INTERNAL.LABELS where name like '%2023-09-08-01-46-00%'
[INFO] SQL stmt to find all the probes: select name from INTERNAL.PROBES where name like '%2023-09-08-01-46-00%'
[INFO] SQL in delete function: select name from INTERNAL.PROBES where name like '%2023-09-08-01-46-00%'
```
will fix that one if we're happy with this workflow change

rymurr (approved):
i think this could be cleaner adn DRYer but happy to do that in a follow pu

joshelser:
> i think this could be cleaner adn DRYer but happy to do that in a follow pu

Yeah, i wasn't happy about all of the copy-paste between jobs. I'll file a note and come back to it.

joshelser:
Talked with Vicky -- this Cypress test failure is latent and non-deterministically failing (at least for me, locally). She has experienced something similar, but thinks this one is slightly different. I'll file a new issue (at her request).

joshelser:
Actually, I'm thinking better of merging this right now. I'd rather get out a new opscenter release and then deal with this after.

Issue #261, opened 2023-09-07
Upgrade from v1.19 to 9ef03e53524247cf9a297317678a0e02a1f83eaa failed


joshelser:
```
File: 042_utils.sql
[Uncaught exception of type 'STATEMENT_ERROR' on line 41 at position 0 : Uncaught exception of type 'STATEMENT_ERROR' on line 4 at position 7 : Uncaught exception of type 'STATEMENT_ERROR' on line 3 at position 0 : SQL compilation error: error line 1 at position 16
invalid identifier 'QUERY_PARAMETERIZED_HASH'] was primary error.  (line 105)
```

rymurr:
so the package is enabled but we haven't updated our table yet. Hard to get around this...maybe we have to skip the proc call here and just call it in finalze_setup

joshelser:
Ahhhh! Because we update the `INTERNAL_REPORTING` views but the `REPORTING` views are reading from the `INTERNAL_REPORTING_MV` tables. I was about to say "I don't get why this fails"

PR #262, opened 2023-09-07
Don't enable query hash until finalize_setup

joshelser:
ref #261

need to test this locally to verify it. finalize_setup() already calls this procedure.


rymurr (approved):

joshelser:
Released v1.19 then upgraded to this commit and things worked as expected. merging.

Issue #263, opened 2023-09-07
query hash columns are misordered


joshelser:
```
our account_usage.query_history has
hash, hash_version, param_hash, param_hash_version.
materialized table has
param_hash, hash, param_hash_version, hash_version (but with correct column data types)
reporting views have
param_hash, hash, param_hash_version, hash_version (with wrong data types)
```

the reporting.enriched_query_history view has jumbled data types assigned to the 4 new columns from Snowflake. It appears that we re-order the columns on the materialized table (while retaining the correct data types) and then combined the materialized table ordering to column names but the original ordering of datatypes from account_usage.query_history to the final reporting.enriched_query_history view.

joshelser:
```
{
  "alter_statement": "ALTER TABLE \"INTERNAL_REPORTING_MV\".\"QUERY_HISTORY_COMPLETE_AND_DAILY_INCOMPLETE\" ADD \"QUERY_HASH_VERSION\" NUMBER, \"QUERY_PARAMETERIZED_HASH\" TEXT, \"QUERY_PARAMETERIZED_HASH_VERSION\" NUMBER, \"QUERY_HASH\" TEXT"
}
```
```
{
  "alter_statement": "ALTER TABLE \"INTERNAL_REPORTING_MV\".\"QUERY_HISTORY_COMPLETE_AND_DAILY\" ADD \"QUERY_PARAMETERIZED_HASH\" TEXT, \"QUERY_HASH\" TEXT, \"QUERY_PARAMETERIZED_HASH_VERSION\" NUMBER, \"QUERY_HASH_VERSION\" NUMBER"
}
```
tracing ftw

joshelser:
Resolved by #265

PR #265, opened 2023-09-07
Fix mis-ordered columns between internal views and staging table

joshelser:
This ensures that the column ordering on the materialization QH table is in the order as the account_usage.query_history view. It would be better to enumerate all columns (rather than try to retain ordering), but also a much larger change.

rel #263

### DIFF
--- a/bootstrap/005_table_migrations.sql
+++ b/bootstrap/005_table_migrations.sql
@@ -25,7 +25,9 @@ BEGIN
       RAISE TABLE_NOT_EXISTS;
   END IF;
 
-  -- Super important that the columns to add are consistently generated, else we may generate conflicting schemas between two tables created from the same view (the internal_reporting_mv query history tables)
+  -- Super important that the columns to add are generated in the same order as the view. This ordering is pivotal for
+	-- the migration algorithm to successfully function and make sure the multiple materialized views in
+	-- INTERNAL_REPORTING_MV have the same schema (and can be UNION'ed together).
   let columns_to_add string := (
       SELECT LISTAGG('"' || COLUMN_NAME || '" ' || DATA_TYPE, ', ') WITHIN GROUP (ORDER BY ORDINAL_POSITION)
       FROM (
@@ -36,13 +38,13 @@ BEGIN
 
       );
 
-  -- Include ordinal_position to fix tables which have the correct columns but are in an unexpected order
+  -- Any columns which exist in the table but not in the view should be dropped, regardless of the ordinal_position.
   let columns_to_drop string := (
       SELECT LISTAGG('"' || COLUMN_NAME || '" ' , ', ')
       FROM (
-      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
+      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
       MINUS
-      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :view_schema AND TABLE_NAME = :view_name
+      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :view_schema AND TABLE_NAME = :view_name
       )
 
       );
@@ -86,19 +88,6 @@ BEGIN
     call INTERNAL.UPDATE_LABEL_VIEW();
     return 'Success';
 END;
-
-create or replace function internal.generate_column_def(source_schema varchar, source_table varchar)
-returns string
-as
-$$
-      -- Ordering the LISTAGG by ORDINAL_POSITION is not strictly necessary, but should eliminate confusion when LISTAGG would
-      -- otherwise generate a random ordering of columns each time it is called.
-      SELECT LISTAGG('"' || COLUMN_NAME || '" ' || DATA_TYPE, ', ') WITHIN GROUP (ORDER BY ORDINAL_POSITION)
-      FROM (
-      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = source_schema AND TABLE_NAME = source_table
-      )
-$$;
-
 
 create or replace function internal.generate_column_names(source_schema varchar, source_table varchar)
 returns string
--- a/bootstrap/005_table_migrations.sql
+++ b/bootstrap/005_table_migrations.sql
@@ -25,22 +25,24 @@ BEGIN
       RAISE TABLE_NOT_EXISTS;
   END IF;
 
+  -- Super important that the columns to add are consistently generated, else we may generate conflicting schemas between two tables created from the same view (the internal_reporting_mv query history tables)
   let columns_to_add string := (
-      SELECT LISTAGG('"' || COLUMN_NAME || '" ' || DATA_TYPE, ', ')
+      SELECT LISTAGG('"' || COLUMN_NAME || '" ' || DATA_TYPE, ', ') WITHIN GROUP (ORDER BY ORDINAL_POSITION)
       FROM (
-      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :view_schema AND TABLE_NAME = :view_name
+      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :view_schema AND TABLE_NAME = :view_name
       MINUS
-      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
+      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
       )
 
       );
 
+  -- Include ordinal_position to fix tables which have the correct columns but are in an unexpected order
   let columns_to_drop string := (
       SELECT LISTAGG('"' || COLUMN_NAME || '" ' , ', ')
       FROM (
-      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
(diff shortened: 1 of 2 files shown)


joshelser:
Talked with Ryan IRL -- he had meant to squash the CTAS which has a `select *` in the QH materialization process and missed it. Rather than bandaid over it, I'll just fix it here.

joshelser:
Reverted my original fix to do `within group (order by ordinal_position)` since we don't actually need it. I think it would be confusing since we have it only in one place and not everywhere.

joshelser:
ugh.

```
Uncaught exception of type 'STATEMENT_ERROR' on line 34 at position 8 : Uncaught exception of type 'STATEMENT_ERROR' on line 5 at position 4 : SQL compilation error: syntax error line 1 at position 1,927 unexpected '00'. syntax error line 1 at position 1,967 unexpected '00'.
```

Not sure where that's coming from (in case I can't figure it out tonight)

joshelser:
> Not sure where that's coming from

Forgot that I had to quote the variables in the where clause since they're no longer bind variables.

bootstrap/019_query_history_updates.sql
```diff
@@ -49,7 +49,10 @@ BEGIN
         end if;
 
         DROP TABLE IF EXISTS RAW_QH_EVT ;
-        CREATE TABLE RAW_QH_EVT AS SELECT * FROM INTERNAL_REPORTING.QUERY_HISTORY_COMPLETE_AND_DAILY WHERE filterts >= :oldest_running AND end_time >= :newest_completed;
```
rymurr, line 52: the warehouse iupdates also has an ctas

joshelser: gosh darn it. I had that thought, but convinced myself (without looking) that we didn't have a staging table for warehouses :(

joshelser:
I've just noticed that while materialization is functioning with this change, the reporting views are broken.

The `create or replace view reporting.enriched_query_history  as select * .. from internal_reporting_mv.query_history_complete_and_daily` is swapping the column names and data types for the query_hash columns.

the internal-reporting_mv:
Screenshot 2023-09-08 at 2 46 45 PM

the reporting view:
Screenshot 2023-09-08 at 2 47 00 PM

I have to assume that we also need to push the column name enumeration into these views.

joshelser:
Thank heavens -- the migrations and views all appear to be working correctly now.

rymurr:
hmmm....from my understanding of snowflake I am not sure how the bug that we are seeing can happen. I don't want to make this fix if we don't have to so I would like to characterise this bug a bit more. The underlying cause of this issue could be very important for how we decide on migrations in the future.

joshelser:
> from my understanding of snowflake I am not sure how the bug that we are seeing can happen

Yeah, the _only_ way I could come up with this happening was that if we were reading column metadata from two different relations and merging name and datatype (incorrectly). Very unexpected.

joshelser:
Thanks to Ryan for prodding on this -- the user-facing `reporting` views had the wrong data types because the ordering on the `INTERNAL_REPORTING_MV.QUERY_HISTORY_COMPLETE_AND_DAILY` and `INTERNAL_REPORTING_MV.QUERY_HISTORY_COMPLETE_AND_DAILY_INCOMPLETE` tables had an equivalent set of columns+types added to them, but in different orders.

Then, we would re-create the views on top of those tables like:
```
    select
        * exclude (record_type)
    from internal_reporting_mv.QUERY_HISTORY_COMPLETE_AND_DAILY
    union all
    select
        * exclude (record_type)
    from internal_reporting_mv.QUERY_HISTORY_COMPLETE_AND_DAILY_INCOMPLETE
```

Thusly, confusing the final view because we now have conflicting orders on the "complete" and "incomplete" tables.

joshelser:
Some cliff notes for 8c4e9bd because it's real tricky. The latest commit does the following:

* Re-introduces the ordering by ordinal_position to `migrate_if_necessary` which ensures that multiple calls from one view to different tables will generate tables with the same ordered schema (not same set of columns/types). This will ensure forward stability as we add new columns.
* * If a table already has the columns from the view but in a different order, we drop those columns and then re-add them in the same order as the view has them.

* * This prevents long-term debt where an OpsCenter user may have a misordering between view and table.

* Adds a check to `refresh_queries(boolean)` which tries to detect if the `QUERY_HISTORY_COMPLETE_AND_DAILY` and `QUERY_HISTORY_COMPLETE_AND_DAILY_INCOMPLETE` tables have the same schema
* * If these two tables do not have the same schema, the "incomplete" table is re-created from the "complete" table (i.e. `create table ... like ..`)

* * This doesn't cause data loss as we re-generate this table every time the procedure runs

* * We then re-create the reporting views now that the schema (the union of complete and incomplete tables) is corrected.
* We don't need a similar "check" for warehouse events because we have not exposed any new columns since the unordered LISTAGG was added. However, the solution to keep incomplete and complete warehouse event tables' schemas in sync will apply as the WH tables use the same `migrate_if_necessary` procedure

I believe this should handle all possible situations where these two query history materialized tables are out of "whack" without forcing a reset. This should let us continue to use the same approach when we want to add the next set of columns.

joshelser:
Just finished testing the upgrade path from v1.19 with this change and verified that the resulting state is correct (materialization task ran, reports have data, etc)

bootstrap/005_table_migrations.sql
```diff
@@ -25,22 +25,24 @@ BEGIN
...
       )
 
       );
 
+  -- Include ordinal_position to fix tables which have the correct columns but are in an unexpected order
   let columns_to_drop string := (
       SELECT LISTAGG('"' || COLUMN_NAME || '" ' , ', ')
       FROM (
-      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
+      SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :table_schema AND TABLE_NAME = :table_name
```
rymurr, line 43: how is the ordinal being use dhere?

bootstrap/005_table_migrations.sql
```diff
@@ -50,16 +52,17 @@ BEGIN
   let alter_table_add_column string := '';
   let alter_table_drop_column string := '';
 
-  if (columns_to_add <> '') then
-    alter_table_add_column := ' ADD ' || columns_to_add;
-    execute immediate alter_statement || alter_table_add_column;
-  end if;
-
+  -- Drop columns first, in case we are re-creating the same columns but in a different order.
```
rymurr, line 55: I am not sure how this works...I also don't think we can drop columns on these tables...thats effectively deleting data

bootstrap/005_table_migrations.sql
```diff
@@ -86,18 +89,24 @@ BEGIN
...
+create or replace function internal.generate_column_list(source_schema varchar, source_table varchar)
 returns string
 as
 $$
-begin
-  let columns string := (
       SELECT LISTAGG('"' || COLUMN_NAME || '"', ', ')
       FROM (
-      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = :source_schema AND TABLE_NAME = :source_table
+      SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = source_schema AND TABLE_NAME = source_table
```
rymurr, line 98: should this be ordered for consistency?

bootstrap/005_table_migrations.sql
```diff
@@ -111,7 +120,40 @@ begin
...
     call internal.generate_insert_statement_cmd(:target_schema, :target_table, :source_schema, :source_table, :where_clause) into stmt;
     execute immediate stmt;
+    SYSTEM$LOG_INFO('Running INSERT with generated query: ' || :stmt);
     let inserted number := (select * from TABLE(RESULT_SCAN(LAST_QUERY_ID())));
   return :inserted;
 end;
 $$;
+
+
+create or replace procedure internal.generate_ctas_statement(target_schema varchar, target_table varchar, source_schema varchar, source_table varchar, where_clause varchar)
```
rymurr, line 130: we don't need ctas anymore do we? We agreed to re-order? Or am I missing something

bootstrap/018_warehouse_updates.sql
```diff
@@ -67,7 +67,10 @@ BEGIN
         end if;
 
         DROP TABLE IF EXISTS RAW_WH_EVT ;
-        CREATE TABLE RAW_WH_EVT AS SELECT * FROM INTERNAL_REPORTING.CLUSTER_AND_WAREHOUSE_SESSIONS_COMPLETE_AND_DAILY WHERE filterts >= :oldest_running AND SESSION_END >= :newest_completed;
+
+        -- Generate the staging table with the enumerated list of columns to avoid a 'select *'
```
rymurr, line 71: the ctas will have the same schema as the underlying view, not sure we need this anymore

bootstrap/005_table_migrations.sql
joshelser: Crud, you're right. I was thinking that we would just re-materialize, but that won't be happening automatically for the unlucky folks who have some materialized data since v1.19.

bootstrap/005_table_migrations.sql
joshelser: Yeah, i just didn't unwind enough commits. CTAS shouldn't need this indirection anymore.

bootstrap/005_table_migrations.sql
joshelser: I didn't add it because it wasn't strictly necessary, but it won't hurt.

joshelser:
393bfab is my next stab at fixing the case where the INTERNAL_REPORTING_MV tables have the correct set of columns but materialized in the wrong order. It tries to detect when the set of columns is the same between the view and the table but the ordering of those columns is wrong.

From your last review @rymurr, the `drop column` and `add column` approach to re-order the columns (the approach which you pointed out was flawed as it caused data loss) was meant to handle one specific case:
* someone installed v1.19
* enabled a bundle that added new QUERY_HISTORY columns
* QH materialization ran and added misordered the columns on the internal_reporting_mv.query_history_complete_and_daily table compared to the 
internal_reporting.query_history_complete_and_daily view (which mirrors snowflake.account_usage.query_history)

I wasn't able to come up with a way to generically handle the following case without rewriting the data. I considered a putting a view on top of the materialization to put it back into the correct order, but that would saddle us with lots of complexity in future migration code as well as incremental materializations. I also thought about trying to detect the subset of columns that were "out of order" and rewrite only those columns, but this felt heavy handed as a bug fix (we should strive to avoid creating this kind of situation with migrations, not build code to optimize for it).

I thought it was best to try to determine if/when one OpsCenter install is subject to this situation and implicitly handle the rewrite without requiring the user to click "Reset". Long term, I think we need to be very pedantic (e.g. clear steps around how the migration change was tested) around changes to 005_table_migrations.sql, but should keep moving forward.

joshelser:
When we were talking yesterday, you had mentioned, @rymurr :

> ... we have already released out of order tables

.. and I agreed with you. However, I don't think that's actually the case after some more inspection.

Your change in #256 fixed a separate bug in OpsCenter where we were not picking up the latest set of columns from `snowflake.account_usage`. This means that, regardless of the set of columns in `snowflake.account_usage.query_history`, OpsCenter was only seeing the set of columns defined in the "faked" opscenter.account_usage.query_history from`003_fake_account_usage.sql`. This means it is impossible for any earlier version of OpsCenter to have misordered columns on the materialized views in `internal_reporting_mv`. The only problem has been our in-development deploys we've used for testing (and the "failed" v1.20 release candidate).

Thus, this migration problem only started after v1.19 and I don't think we need the destructive rebuild that I made in 393bfab. Testing this out to be certain before I revert it from this PR.

rymurr (approved):

PR #274, opened 2023-09-07
248: Add loading check method

susannahmarcus:
- tested on first queries test
- gitIgnore node modules and yarn.lock
- add loading check method to error check method
- prettier ran
- changes the config to add more options


joshelser (dismissed):
LGTM. Should we wait for Vicky to also review?

cypress/support/utils.js
```diff
@@ -62,66 +58,60 @@ export const fillInProbeForm = (
...
+export const fillInNewUngroupedLabelForm = (labelName, condition) => {
+  if (labelName != "None") {
+    cy.get('input[aria-label="Label Name"]').clear().type(labelName);
   }
 
-  if ( condition != "None" ) {
+  if (condition != "None") {
     cy.get('textarea[aria-label="Condition"]')
       .clear()
-      .type(condition).type('{command+enter}');
```
vqmarkman, line 124: @susannahmarcus, this one should be replaced back to be `type('{command+enter}')`.
Broke tests for me in the past.

susannahmarcus: To the best of my knowledge, this should be fine in that formatting since it just spaced it over two lines, but from a JS perspective, this is the same thing

susannahmarcus: that said, I need to update something that I'm seeing is failing on deploy and see if that helps...

susannahmarcus:
I added something to the clickCheck function which should make it work, despite rerenders (which seems to be a problem in the probes test, so this would help that function in that instance as well)

cypress/support/utils.js
vqmarkman: this is my bad, I thought it replaced **command+enter** with **command**
we are good! Sorry about that.

vqmarkman (approved):
Ship it. Thank you, Susannah!

vqmarkman:
I don't like probes test failure - it is very consistent which indicates we have a problem.
Lemme look at it first before you merge.

PR #289, opened 2023-09-08
Exclude ENABLED from CATALOG.LABELS

### LINKED ISSUE #286: ENABLED column is null for label tables
Looks like we forgot to set the new ENABLED column in MERGE that run over internal.labels and interanl.predefined_labels.

joshelser:
We aren't setting this field and provide no means to change a label being enabled or not. Hide it for now.

Closes #286

### DIFF
--- a/bootstrap/004_labels.sql
+++ b/bootstrap/004_labels.sql
@@ -19,6 +19,9 @@ BEGIN
         UPDATE INTERNAL.LABELS SET IS_DYNAMIC = FALSE WHERE IS_DYNAMIC IS NULL;
     END IF;
 
+    -- Set ENABLED=TRUE for all labels, until we have a use-case where labels can be disabled.
+    update internal.labels set enabled = true where enabled is null;
+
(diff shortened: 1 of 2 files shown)


joshelser:
> why remove it? Why not just set it? Before removing it I would like to understand why it was there in the first place.

We don't have a user-facing notion of a label being enabled or not which is why I thought it better to remove it. We do have an internal concept in that some labels depend on a snowflake bundle being enabled (e.g. query_hash).

I thought about setting `enabled=true` (thinking that we would eventually making labels which could be enabled/disabled), but then figured we should add that with specific intention.

rymurr:
Totally agree, makes a lot of sense. Thanks for clarifying.

I am debating if its harder to maintain long term if we have that except clause floating around? I think we shold at least put in a comment for our future selves. So I levaes it toyou: comment this and go with it or set to true by default

joshelser:
I went back to the history and found that we had the enabled column on labels from day 1 -- considering that, it felt more like a bug, so I updated the code to assume `enabled =true` for label operations and set all `enabled is null` to be `true`.

rymurr (approved):

Issue #270, opened 2023-09-07
UI tests: add new loading method into all settings.spec.cy tests


susannahmarcus:

susannahmarcus:
addressed in issue #290

Issue #252, opened 2023-09-06
UI tests: fix probes and labels tests where element can't be found in case of multiple label groups


vqmarkman:
Can be observed here: https://github.com/sundeck-io/OpsCenter/actions/runs/6097333475/job/16546848806

It looks like the problem was with multiple tabs on labels page where one of the runs left labels not deleted at the end of the run and subsequent run was barfing on it.

Dropping database at the end of the cypress run resolved the issue. This is a test issue.

Issue #40, opened 2023-07-12
Clicking left sidebar for labels will return user to last visited page


joshelser:
The sidebar appears to not operate like we would expect for a "normal website". For example:

* Visit "Labels" page
* Click "New"
* Click to the "Probes" page in sidebar
* Click back to "Labels" page in sidebar
* Find yourself on "New label" page and not the "list of labels" page

This is probably something to do with Streamlit remembering what "sub page" the user was last on. Maybe it is something we can control/influence.

PR #290, opened 2023-09-09
Issues #252, #268, #269, #270, #271

vqmarkman:
Issue #268: new method of confirmation that page was loaded.
Issue #252: fix probes and labels tests where element can't be found in case of multiple label groups

### DIFF
--- a/cypress/e2e/warehouse.spec.cy.js
+++ b/cypress/e2e/warehouse.spec.cy.js
@@ -45,11 +45,11 @@ describe("Warehouse section", () => {
     }
   });
 
-  // Test is skipped due to a known issue: exception while iterating over ranges
-  // pandas/core/indexes/range.py", line 349, in get_loc
-  it.skip("Menu: Warehouses (Heatmap)", () => {
+  it("Menu: Warehouses (Heatmap)", () => {
 
     cy.visit("/");
+
+    checkForLoading();
 
     const stringList= ["365", "90", "30", "7"];
 
@@ -59,16 +59,11 @@ describe("Warehouse section", () => {
       .should("be.visible");
 
     // Click on Filters
-    cy.get('div[data-testid="stMarkdownContainer"]')
-      .contains("Filters")
-      .click();
+    clickCheck({ clickElem: 'div[data-testid="stMarkdownContainer"]', contains: "Filters" });
 
     for (const str of stringList) {
-      cy.get('button[kind="secondary"]')
-        .contains(str)
-        .click();
+      clickCheck({ clickElem: 'button[kind="secondary"]', contains: str });
 
-      checkNoErrorOnThePage();
     }
   });
 
--- a/cypress/e2e/warehouse.spec.cy.js
+++ b/cypress/e2e/warehouse.spec.cy.js
@@ -1,4 +1,8 @@
-import { checkNoErrorOnThePage, setup } from "../support/utils";
+import { checkNoErrorOnThePage,
+         setup,
+         checkForLoading,
+         clickCheck, }
+from "../support/utils";
 
 describe("Warehouse section", () => {
   before(() => {
@@ -9,14 +13,11 @@ describe("Warehouse section", () => {
 
     cy.visit("/");
 
-    cy.wait(10000);
+    checkForLoading();
 
     const stringList= ["365", "90", "30", "7"];
 
-    cy.get("span")
-      .contains("Warehouses")
-      .should("be.visible")
-      .click();
+    clickCheck({ clickElem: "span", contains: "Warehouses" });
 
     // Check that text "Warehouse Heatmap" is visible
     cy.get("span")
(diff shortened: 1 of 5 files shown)


vqmarkman:
@susannahmarcus, I've implemented all loading solution for all of the reports.

**Remaining issues:**
- issue 288: queries report is randomly failing in unclear place
- issue 40: Clicking left sidebar for labels will return user to last visited page #40  (or I think it is the same issue)

My suggestion would be: let's fix everything we know about in this PR, make it stable and merge.
It seems to me that we are very close.

cypress/e2e/warehouse.spec.cy.js
```diff
@@ -29,32 +30,27 @@ describe("Warehouse section", () => {
...
-    cy.get('li[role="option"]')
-        .should("be.visible")
+    clickCheck({ clickElem: 'li[role="option"]', contains: "Warehouse Activity" });
+
+    // Check that text "Warehouse Activity" is visible
+    cy.get("span")
         .contains("Warehouse Activity")
-        .should("be.visible")
-        .click();
+        .should("be.visible");
```
susannahmarcus, line 38: there seems to be a click that was taken out here

cypress/support/utils.js
```diff
@@ -131,13 +128,16 @@ export const fillInNewGroupedLabelForm = (
...
 export const addNewLabelToGroup = (groupName, labelName, condition, rank) => {
+
   // Find tab with the group name and click on it
   cy.get('button[data-baseweb="tab"')
     .should("exist")
     .find('div[data-testid="stMarkdownContainer"]')
     .should("exist")
     .contains(groupName)
-    .click();
+    .as("labelGroupTab");
```
susannahmarcus, line 138: we might want to refactor this to use the as() earlier. sometimes, by the time the test reaches contains() there has been a rerender, so it's better to make an alias earlier on. I haven't looked at these tests yet, but if this point is flaky then that could be one thing we could try.

cypress/support/utils.js
```diff
@@ -170,6 +171,17 @@ export const addNewLabelToGroup = (groupName, labelName, condition, rank) => {
 
 // For ungrouped label, specify "Ungrouped" in the groupName argument
 export const labelDelete = (groupName, labelName) => {
+  cy.log("*** labelDelete (begin): groupName:labelName", groupName,labelName);
```
susannahmarcus, line 174: nit: do we want to keep this log?

cypress/support/utils.js
```diff
@@ -181,8 +193,10 @@ export const labelDelete = (groupName, labelName) => {
       cy.get('div[data-testid="column"]')
         .contains("🗑️")
         .should("exist")
-        .click();
+        .click({ force: true });
     });
+
+  cy.log("*** labelDelete (end): groupName", labelName);
```
susannahmarcus, line 199: nit: do we want to keep this log?

cypress/support/utils.js
```diff
@@ -294,9 +308,7 @@ export const clickCheck = (options) => {
 
 export const checkForLoading = () => {
   cy.get('[data-testid="stMarkdownContainer"]')
-    .contains("Please wait...")
-    .should("not.exist", { timeout: 120000 });
-  cy.get('[data-testid="stStatusWidget"]').should("not.exist", {
-    timeout: 120000,
-  });
+    .contains("Please wait...", { timeout: 360000 })
```
susannahmarcus, line 311: i _think_ we might want to move the timeout to the "get" not the contains. I'll check it locally once merged

susannahmarcus (approved):
there aqre some things we might wanna fix, but I'll look into them locally post merge, when I'm rebasing onto these changes

cypress/e2e/warehouse.spec.cy.js
vqmarkman: yea, the diff shows it in a weird way. There was a click that I replaced with a call to clickCheck() and then I follow up with checking if we can actually see "Warehouse Activity" (that might be redundant actually, but not harmful ... I think)

cypress/support/utils.js
vqmarkman: yes, you are absolutely correct! I've already refactored couple of calls like this for probes I believe.

cypress/support/utils.js
vqmarkman: will keep it for now while we are stabilizing things (easier to read cypress output)

cypress/support/utils.js
vqmarkman: will remove after we are done debugging.

cypress/support/utils.js
vqmarkman: not sure ... I felt that it needed to be waiting for "Please wait" to appear ... but we will see after the merge.
Thanks for the thorough review, @susannahmarcus ! Really appreciate it.

PR #298, opened 2023-09-13
Two more migration fixes

joshelser:
1. Truncate the predefined_labels table before we re-define the predefined labels.
2. Don't recreate reporting views before MV migration

Doing an upgrade test now.


rymurr (dismissed):

joshelser:
> Doing an upgrade test now.

Successfully upgraded from v1.19 with this change, but the new predefined labels don't show up as user labels because we're truncating the table (current hunch)

joshelser:
> the new predefined labels don't show up as user labels because we're truncating the table (current hunch)

This is a wholly separate issue, it seems where predefined labels are not made visible to users in the upgrade path. I want to separate this from the view schema migration bugfixes.

Merging this to keep moving forward.

PR #304, opened 2023-09-13
273 & 300: Add typescript to cypress test env and increase viewport

susannahmarcus:
updated utils to use typescript objects for their args, have viewport matching e2e test for other cypress env, and skipped that reset test since it's costly and takes too long so it times out every time.


cypress/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "lib": [
+      "es5",
+      "dom"
+    ],
+    "target": "es5",
```
doronrosenberg, line 7: why es5 and not es2020?

doronrosenberg (approved):

PR #311, opened 2023-09-14
Reduce size of reoccurring query report

### LINKED ISSUE #310: Reduce size of query hash report table
Streamlit on snowflake is limted to 2MB per message and the query has report table was much larger (some very long queries) this produced an ugly error messag.

rymurr:
closes #310

### DIFF
--- a/app/ui/reports_query_hash.py
+++ b/app/ui/reports_query_hash.py
@@ -80,12 +80,14 @@ select sum(cost) as "Cost", sum(cnt) as "Count", case when "IsRepeated" then 'Re
 
     def top_table():
         val = st.radio(
-            "Top 1000 repeated queries by:",
+            "Top 100 repeated queries by:",
             ["Count", "Cost", "Cost per Query"],
         )
         sql = f"""
             with raw as (
-            select any_value(query_text) as "Query Text", to_varchar(sum(cost), '999,999.00') as "Cost", count(*) as "Count", to_varchar(sum(cost)/count(*), '999,999.000000') as "Cost per Query"
+            select any_value(query_text) as "Query Text", any_value(query_id) as "Query Id",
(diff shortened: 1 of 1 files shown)


app/ui/reports_query_hash.py
```diff
@@ -105,7 +107,14 @@ def top_table():
...
         """
         )
-        st.dataframe(df, use_container_width=True)
+        df["Query Text"] = df.apply(
+            lambda row: f"{row['Query Id']}: {row['Query Text'][:10000]}"
+            if len(row["Query Text"]) > 10000
+            else row["Query Text"],
+            axis=1,
+        )
+        print(len(df.to_json()))
```
joshelser, line 116: remove

app/ui/reports_query_hash.py
```diff
@@ -105,7 +107,14 @@ def top_table():
 
         """
         )
-        st.dataframe(df, use_container_width=True)
+        df["Query Text"] = df.apply(
+            lambda row: f"{row['Query Id']}: {row['Query Text'][:10000]}"
```
joshelser, line 111: Regular python `str`'s would do the correct thing if you did `"asdf"[:1000]` and the length of the string was less than the offset value. Does the dataframe act differently and throw some ArrayIndexOutOfBounds?

joshelser (changes requested):

app/ui/reports_query_hash.py
rymurr: Row is a dict and the query text field is a str. Should behave just fine.

joshelser: I missed the first half of the f-string in the if-clause. I thought it was just doing:
```
if len(queryText) > 1000:
  queryText[:1000]
else:
  queryText
```

joshelser (approved):

PR #312, opened 2023-09-14
Reduce area between rows of probes

rymurr:
Weird interaction with Streamlit, the `div` wants to show the label even though its default hidden.

Before:

After:


joshelser (approved):
how silly.

PR #316, opened 2023-09-14
Missed making the snowsql config in cleanup

joshelser:
amends #260

noticed the failure in https://github.com/sundeck-io/OpsCenter/actions/runs/6191071962/job/16808966878?pr=315, just forgot to do the copy-paste in cleanup.


joshelser:
Opened #317 to consolidate into an action which would have prevented this.

rymurr (approved):

PR #318, opened 2023-09-14
Disable gatherUsageStats and runner.fastReruns

joshelser:
The first one will remove all of the api.segment.io calls in the cypress output and the second one turns off that fun feature where Streamlit will allow navigation to a new page while the current one is still loading.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -79,7 +79,7 @@ jobs:
       - name: run deploy
         run: "poetry run python deploy/devdeploy.py -p opscenter"
       - name: run streamlit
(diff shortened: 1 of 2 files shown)


joshelser:
> the second one turns off that fun feature where Streamlit will allow navigation to a new page while the current one is still loading.

Happened to re-read the options on `streamlit run` and this one jumped out at me this time. Naively poking around, it seems to do what we had long hoped Streamlit would do.

vqmarkman:
@susannahmarcus , you have to see this commit!!! Our nightmare might be over.

susannahmarcus (approved):
lgtm 👍

Issue #234, opened 2023-09-04
Refactor labels and probes crud


rymurr:
there is a lot of near duplicate code in labels and probes. We should consolidate this code as CRUD will be a common pattern as we go further

rymurr:
I am going to jump on this @joshelser . Hope you don' tmind. Its the most urgent ticket and I am low on work. I think you are closer to the migration/release stuff and would like you to focus on that.

joshelser:
Keeping a list of things mentioned in code review, but aren't particularly urgent

- [ ] Create an decorator instead of using the `with_session` `@contextmanager` function for passing the snowpark Session into the CRUD module 
- [ ] Create Makefile or similar for devdeploy.py

PR #319, opened 2023-09-15
CRUD for Labels in UI

joshelser:
Migrated from pydantic 2.3 -> 1.10. Cleaned up a bunch of stuff to get create_label working with local streamlit in devdeploy.

Wanted to ask some questions about some specific stuff, will leave comments inline.


app/ui/crud
joshelser, line 1: This is a symlink up and over to crud.

This shouldn't be necessary for the "real" deployment (since we copy all of it into the same app package). Can you think of any better way to make this work for devdeploys (e.g. maybe changing pythonpath)?

app/ui/labels.py
```diff
@@ -174,14 +181,27 @@ def on_create_click(self, name, group, rank, condition, is_dynamic):
...
-            outcome = self.snowflake.call(
-                "ADMIN.CREATE_LABEL", name, group, rank, condition, is_dynamic
-            )
-
-            if outcome is None:
-                self.session.set_toast("New label created.")
-                self.session.do_list()
-                return
+            try:
+                with with_session(self.snowflake) as sf:
```
joshelser, line 185: I need to rename the function to avoid the double "with", but what are you thoughts about this approach to the `context={..}` we could have used with pydantic 2.3?

app/crud/session.py
```diff
@@ -0,0 +1,16 @@
...
+@contextmanager
+def with_session(session):
+    """
+    Makes the give Snowpark Session available to the CRUD implementation for the scope of this call.
+    :param session:
+    :return:
+    """
+    token = session_ctx.set(session)
+    yield session
+    session_ctx.reset(token)
```
rymurr, line 16: nit newline

app/ui/labels.py
rymurr: i like it

app/ui/crud
rymurr: I would be inclined to change pythonpath over this. But its a nit really

app/ui/labels.py
rymurr: I want to try and add this as an annotation so we don't have to manually do the `with` but thats cosmetic only

deploy/devdeploy.py
```diff
@@ -21,13 +23,43 @@ def _setup_database(cur, database: str, schema: str, stage: str):
     )
 
 
+def _make_crud_zip(source: str, dest: str):
```
rymurr, line 26: this makes me think its time to switch to a makefile...not now but this script is becoming complex

rymurr (commented):
looks great!

app/ui/labels.py
joshelser: Great, thanks! I added a note for the decorator in #234 to pick up later.

deploy/devdeploy.py
joshelser: Wrote it down in #234 to not forget for later, will spin into a new issue when we're done with this (I expect we'll have more stuff fall out that we want to do later)

joshelser:
Some top level comments/tasks before this is ready to merge:

- [ ] fix TODO where I just noticed label update is broken
- [ ] make sure new crud pytests are passing (automate in a second step)

app/crud/common.py
```diff
@@ -0,0 +1,63 @@
...
+## TODO
+# migrate - maybe we don't move out of sql?
+# copy predefined labels to labels - maybe we dont move out of sql?
+# same again for probes
+# hook up to stored procs for (CRUD and validation)
+# more tests
+_TYPES = {"LABEL": Label, "PREDEFINED_LABEL": PredefinedLabel}
+
+
+def create_entity(session, entity_type, entity):
```
joshelser, line 14: You'll notice that the begin/commit-rollback is no longer here.

Long-form answer: I pushed it down into the BaseModel class. The original patch had the entire CRUD call (including updating the view after the fact) wrapped in the transaction. This fails when we try to re-create the view after doing the CRUD work. Pushing it down lets us do all of the work for CRUD, commit, and then update the view.

This also has the benefit of converging the streamlit code path with the quickly incoming ADMIN procedures path.

app/crud/labels.py
```diff
@@ -0,0 +1,271 @@
...
+                """
+                SELECT COUNT(*) FROM INTERNAL.LABELS
+                WHERE (NAME = ? AND GROUP_NAME IS NULL) OR (GROUP_NAME = ? AND GROUP_NAME IS NOT NULL)""",
+                params=(
+                    name,
+                    name,
+                ),
+            ).collect()[0][0]
+            # TODO updating a label fails after adding this assertion
+            assert count == 0, "Duplicate label name found. Please use a distinct name."
```
joshelser, line 198: This incorrectly warns on update that you are duplicating a label name. Need to fix.

app/crud/base.py
```diff
@@ -0,0 +1,122 @@
...
+            f"CREATE TABLE IF NOT EXISTS internal.{cls.table_name} ({cols_str})"
+        ).collect()
+        if with_catalog_view:
+            session.sql(
+                f"CREATE OR REPLACE VIEW catalog.{cls.table_name} AS SELECT * FROM internal.{cls.table_name}"
+            ).collect()
+
+    def write(self, session):
+        try:
+            session.sql("BEGIN").collect()
```
rymurr, line 47: can you mvoe the transaction stuff back to a common module? This (for me) mixes business logic w/ util methods and is hard to read. Or was there a good reason to remove it

app/crud/base.py
```diff
@@ -0,0 +1,122 @@
...
+            df.write.mode("append").save_as_table(f"INTERNAL.{self.table_name}")
+            session.sql("COMMIT").collect()
+        except Exception:
+            session.sql("ROLLBACK").collect()
+            raise
+
+        session.call(self.on_success_proc)
+
+    def get_id(self) -> str:
+        return None
```
rymurr, line 58: nit: can you add pydoc? I know I wrote it so I feel a bit embarassed asking

app/crud/base.py
```diff
@@ -0,0 +1,122 @@
...
+            session.sql(
+                f"DELETE FROM INTERNAL.{self.table_name} WHERE {self.get_id_col()} = ?",
+                params=(self.get_id(),),
+            ).collect()
+            session.sql("COMMIT").collect()
+        except Exception as e:
+            session.sql("ROLLBACK").collect()
+            raise e
+
+        session.call(self.on_success_proc)
```
rymurr, line 75: could we instead override the `delete` and call the success proc in the child class?
eg in Labels
```
def delete(self, session):
    super().delete(session)
    session.call(self.on_success_proc)
```

app/crud/base.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def update(self, session, obj) -> "BaseOpsCenterModel":
+        cols = dict(obj)
+        # Filter out `None` values
```
rymurr, line 79: why?

app/crud/common.py
rymurr: dont mind moving it inside but it seems we are re-using the logic and that could still live in a context manager

app/crud/labels.py
```diff
@@ -0,0 +1,271 @@
...
+from pydantic import (
+    validator,
+    root_validator,
+)
+from typing import Optional, ClassVar
+import datetime
+from .base import BaseOpsCenterModel
+from .session import session_ctx
+
+## TODO
```
rymurr, line 12: I think this is duped from an earlier clas

app/crud/labels.py
```diff
@@ -0,0 +1,271 @@
...
+# migrate - maybe we don't move out of sql?
+# copy predefined labels to labels - maybe we dont move out of sql?
+# same again for probes
+# hook up to stored procs for (CRUD and validation)
+# more tests
+# ensure create, prepopulate, validate get called at the right times in bootstrap
+class Label(BaseOpsCenterModel):
+    table_name: ClassVar[str] = "LABELS"
+    on_success_proc: ClassVar[str] = "INTERNAL.UPDATE_LABEL_VIEW"
```
rymurr, line 21: ahh, it looks like you are already calling this in the subclass...why have both the on_success_proc and the lines (39, 43, 66)?

app/crud/labels.py
```diff
@@ -0,0 +1,271 @@
...
+        assert isinstance(time, datetime.datetime)
+        return time
+
+    @validator("enabled", "is_dynamic")
+    def enabled_or_dynamic(cls, b: bool) -> bool:
+        assert isinstance(b, bool)
+        return b
+
+
+class PredefinedLabel(Label):
```
rymurr, line 223: prob should remove this class for now...

deploy/devdeploy.py
```diff
@@ -21,13 +23,47 @@ def _setup_database(cur, database: str, schema: str, stage: str):
...
```
rymurr, line 26: I think we have to do this in deploy also...maybe it should go in helpers?

rymurr (commented):
looking good. Just a few things for cleanup I think

app/ui/crud
```diff
@@ -0,0 +1 @@
+../crud/
```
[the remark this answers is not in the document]
joshelser: Yeah, the zip is there already. I do need to test out if this symlink breaks deploy (my hope is that it's just an artifact for devdeploy that we can ignore the rest of the time). Thanks for the reminder.

app/crud/base.py
joshelser: Yep! I was trying to avoid the duplication of this in the child class by doing this in the parent, but I don't feel strongly.

app/crud/base.py
```diff
@@ -0,0 +1,122 @@
+from snowflake.snowpark import Row
+from pydantic import BaseModel
+from typing import ClassVar, get_args, get_origin, Union, Dict
+import datetime
+
+
+## TODO
```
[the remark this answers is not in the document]
joshelser: Yeah, i've been keeping it up to date. I can probably scratch off the "write tests" part, too.

app/crud/base.py
joshelser: Pushing it down was to keep the two callers (app/ui/labels.py and 004_labels.sql procedures that will be coming) in sync. I could also lift the transaction back up into common.py and have the streamlit code call that common function.

app/crud/base.py
joshelser: Early on, I was getting "none" strings in the database. That might have been bugs later fixed. Let me revisit with some tests.

app/crud/labels.py
joshelser: > why have both the on_success_proc and the lines (39, 43, 66)

I meant to remove it from the Label class and forgot, heh. Intention was to have child classes declare what their "on success" procedure is, and have the parent class automatically call it for them.

app/crud/common.py
joshelser: Got it.

app/crud/base.py
joshelser: 👍🏼  no problem

app/crud/base.py
joshelser: Moved the transaction `@contextmanager` method into `base.py` and updated the concrete classes to keep the view-regeneration outside of the txn.

app/crud/base.py
joshelser: Not an issue anymore

joshelser:
@rymurr I got deploy.py updated, but I'm not entirely sure at first glance how to deploy a Zip (with our third-party python module) and make that available to streamlit.

Plotly and sqlglot both exist on the snowflake conda repo which, I think, is the magic with lets us be able to add them to our environment.yml. Adding in "crud" to environment.yml fails in the expected way

```
Error starting Streamlit: SQL compilation error: Cannot create a Python function with the specified packages. Please check your packages specification and try again. 'Packages not found:

crud'. Hint: These packages are available in other python runtimes:["pydantic=="->[3.9, 3.10, 3.11], "plotly=="->[3.9, 3.10, 3.11], "altair=="->[3.9, 3.10, 3.11], "snowflake-snowpark-python=="->[3.9, 3.10]].
```

Totally expected -- how would Snowflake know that we put a Zip file with a python module in our stage.

To make this work, I feel like we would have to copy the CRUD module into the rest of the streamlit app _and_ copy the zip (for the next PR where we update the procedures to use the new label CRUD functions). Just wanted to double check I'm not missing some dark-corner of native apps.

joshelser:
> To make this work, I feel like we would have to copy the CRUD module into the rest of the streamlit app and copy the zip (for the next PR where we update the procedures to use the new label CRUD functions). Just wanted to double check I'm not missing some dark-corner of native apps.

Yeah, this is kind of annoying! I wish we had the ability to pass the zip file to a streamlit app like we do a procedure.

It feels like we either have to maintain the `app/crud` structure and automate copying the files into `app/ui` and try to test against name conflicts, or we build the zip based on a subset of files in `app/ui` and try to test against any streamlit API leaking into the module.zip.

I'm not sure which is best, but I just tested out the first approach enough to see it work. I'll push that tonight, but will think about it some more and we can catch up tmrw.

rymurr:
> > To make this work, I feel like we would have to copy the CRUD module into the rest of the streamlit app and copy the zip (for the next PR where we update the procedures to use the new label CRUD functions). Just wanted to double check I'm not missing some dark-corner of native apps.
> 
> Yeah, this is kind of annoying! I wish we had the ability to pass the zip file to a streamlit app like we do a procedure.
> 
> It feels like we either have to maintain the `app/crud` structure and automate copying the files into `app/ui` and try to test against name conflicts, or we build the zip based on a subset of files in `app/ui` and try to test against any streamlit API leaking into the module.zip.
> 
> I'm not sure which is best, but I just tested out the first approach enough to see it work. I'll push that tonight, but will think about it some more and we can catch up tmrw.

I think we either have to flatten `app/crud` or we have to figure out how to reference the zip file. We may simply be able to put the zipfile in the stage next to the streamlit app: https://realpython.com/python-zip-import/

Can talk more when you are awake

rymurr:
> Plotly and sqlglot both exist on the snowflake conda repo which, I think, is the magic with lets us be able to add them to our environment.yml. Adding in "crud" to environment.yml fails in the expected way

sqlglot doesnt exist in their conda repo (or at least not last I checked) we refer to that directly in stored procs. We don't need to refer to it in `environment.yml` if it has no dependencies and is available elsewhere (eg through the stage)

rymurr (commented):
looks great, 1 nit then we have to sort out the zip/deploy stuff

app/ui/Home.py
```diff
@@ -52,5 +51,5 @@
     st.container()
 
     filter_values = filters.display(filter_container)
-    with st.spinner("Loading Warehouse Heatmap"):
-        reports_heatmap.heatmap(filter_values, credit_cost)
+    # with st.spinner("Loading Warehouse Heatmap"):
```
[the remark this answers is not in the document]
joshelser: nope, thanks for catching. I got annoyed at how slow it was to load the home page yesterday.

joshelser:
> sqlglot doesnt exist in their conda repo (or at least not last I checked)

Sqlglot (11.5.7) is in their conda repo now! Though, apologies, I don't know what I saw that made me think that we were using sqlglot in the streamlit app (I must have gotten plotly and sqlglot switched in my head).

> I think we either have to flatten app/crud or we have to figure out how to reference the zip file. We may simply be able to put the zipfile in the stage next to the streamlit app: https://realpython.com/python-zip-import/

Excellent, this sounds like the best option so far. Let me see if I can hack something together.

deploy/devdeploy.py
```diff
@@ -23,12 +24,23 @@ def _setup_database(cur, database: str, schema: str, stage: str):
...
         cur.execute(put_cmd)
 
+    # Also copy the CRUD zip into @stage/ui because the streamlit app cannot reach "out" of the `ui` directory
+    # to dynamically load this from elsewhere in the stage.
+    put_cmd = f"PUT 'file://app/python/crud.zip' '@{schema}.{stage}/ui' overwrite=true auto_compress=false"
+    print(put_cmd)
+    cur.execute(put_cmd)
+
+    # Copy the CRUD zip from app/python to app/ui to keep devdeploy and deploy streamlit consistent with each other.
+    shutil.copy2('app/python/crud.zip', 'app/ui/crud.zip')
```
joshelser, line 42: @rymurr curious what you think here. We have a couple of ways we could make the crud module available to the streamlit app for local dev.

* Set `PYTHONPATH` when doing `run streamlit app/ui/Home.py`
* Use the symlink `ln -s ../crud app/ui/crud`
* Let devdeploy.py rebuild the zip and use the zip (only) in the app

The last one is what this PR is doing because it keeps us from having to do "are we running locally or inside a native app" check to conditionally load the `crud.zip` module. Any strong opinions for v1?

app/crud/base.py
```diff
@@ -0,0 +1,117 @@
...
+        ).collect()
+
+    def update(self, session, obj) -> "BaseOpsCenterModel":
+        cols = dict(obj)
+        # Build up the SET clause and bind param values
+        set_elements = []
+        params = []
+        for k, v in cols.items():
+            set_elements.append(f"{k} = ?")
+            params.append(v)
```
joshelser, line 71: I'm not sure why this is happening, but I get an error running inside of Snowflake but not when running locally. Feels like some difference in version of snowpark session API.

```
Numeric value 'None' is not recognized
File "/usr/lib/python_udf/7529d9b466ef006af9b8b8149d6967acf5f82bbeb9479eea064c6239760d9a0a/lib/python3.8/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 561, in _run_script
    self._session_state.on_script_will_rerun(rerun_data.widget_states)
File "/usr/lib/python_udf/7529d9b466ef006af9b8b8149d6967acf5f82bbeb9479eea064c6239760d9a0a/lib/python3.8/site-packages/streamlit/runtime/state/safe_session_state.py", line 68, in on_script_will_rerun
    self._state.on_script_will_rerun(latest_widget_states)
File "/usr/lib/python_udf/7529d9b466ef006af9b8b8149d6967acf5f82bbeb9479eea064c6239760d9a0a/lib/python3.8/site-packages/streamlit/runtime/state/session_state.py", line 482, in on_script_will_rerun
    self._call_callbacks()
File "/usr/lib/python_udf/7529d9b466ef006af9b8b8149d6967acf5f82bbeb9479eea064c6239760d9a0a/lib/python3.8/site-packages/streamlit/runtime/state/session_state.py", line 495, in _call_callbacks
    self._new_widget_state.call_callback(wid)
File "/usr/lib/python_udf/7529d9b466ef006af9b8b8149d6967acf5f82bbeb9479eea064c6239760d9a0a/lib/python3.8/site-packages/streamlit/runtime/state/session_state.py", line 247, in call_callback
    callback(*args, **kwargs)
File "/home/udf/27672521534107002/labels.py", line 236, in on_update_click
    _ = old_label.update(sf, new_label)
File "/home/udf/27672521534107002/crud.zip/crud/labels.py", line 108, in update
    super().update(session, obj)
File "/home/udf/27672521534107002/crud.zip/crud/base.py", line 74, in update
    session.sql(
...
```
feels like snowpark session is trying (and failing) to coerce `None` to an `int`. Updating a label is broken as a result. Trying to figure out what is happening.

joshelser:
Cypress is failing because we're running the app from the local directory which expects to find the crud.zip in app/ui. However, we have a fresh checkout because we're doing devdeploy in one step and then running cypress in another (which doesn't share teh same checkout). I'm going to put a bandaid in to the workflow for now.

Issue #322, opened 2023-09-15
Probes test is consistently failing after we started running with --runner.fastReruns false


vqmarkman:
Issue #1 is that test now throws a warning: `Calling st.experimental_rerun() within a callback is a no-op.` and spits errors on streamlit console:

`2023-09-15 09:49:42.889 `label` got an empty value. This is discouraged for accessibility reasons and may be disallowed in the future by raising an exception. Please provide a non-empty label and hide it with label_visibility if needed.
`

This makes test fail as it does not expect this behavior.

https://github.com/sundeck-io/OpsCenter/assets/104043363/0f2f919a-144d-4cdc-8b27-94842f9cdad4

Did some digging on this topic, and found this link: https://github.com/streamlit/docs/issues/422

PR #324, opened 2023-09-15
Reenable fastReruns as it breaks a probes cypress test somehow

joshelser:
rel #322


vqmarkman:
@joshelser, the very same test that I mentioned in #322 failed in this PR :( 
I don't understand it ... at all. It was incredibly consistent when I ran locally.

vqmarkman (approved):

PR #328, opened 2023-09-19
299: dechunk cypress tests labels

susannahmarcus:
In this pr:
- label tests broken into more digestible chunks
- label form fields are in consistent locations and with consistent naming (Rank v Group Rank)
- added dynamic group labels tests to the label tests
- increased global time out (was able to pass rigorous testing that way)
- shows exception text in console when there is a _specific type of exception that is noted as an exception_. There is another type that just registers as a notification, so that hasn't been added yet...


app/ui/labels.py
```diff
@@ -222,6 +222,10 @@ def create_label(self, grouped: str, is_dynamic: bool):
         st.title("New Label")
         group = None
         rank = None
+        name = None
+
+        if not is_dynamic:
+            name = st.text_input(key="NAME", label="Label Name")
```
joshelser, line 228: I don't think we have strict guidance on this, but what was your thinking with this change?

For grouped labels, I think of it as a hierarchy: `group > label`. It makes more sense to me, when we're creating a grouped label that group comes first, and then name. WDYT?

joshelser (commented):
One question on the label create form changes, but the other changes make sense to me

app/ui/labels.py
vqmarkman: Agree with @joshelser, it does not feel natural to have label first. Especially when the second label in the group is created and group name is greyed out ... looks and feels weird.

cypress/e2e/Labels/tests/multipleLabelsCRUDTests.ts
```diff
@@ -0,0 +1,179 @@
...
+          });
+          checkLabelExists({
+            labelName: label,
+            groupName: UNGROUPED,
+            doesExist: false,
+          });
+        }
+      });
+
+      it("Create multiple ungrouped labels", () => {
```
vqmarkman, line 55: This test does not clean labels after creation. Is it ok? Am I missing some other place where it is done ?

vqmarkman:
Absolutely fabulous tests, Susannah. Will be so much easier to add new and modify them.

This PR also closes: #307 - UI tests for for dynamic labels CRUD

Last question: we had set of negative ungrouped label tests commented out due to a bug that is most likely is already fixed. I don't see them in this PR.  Just want to make sure we did not throw them away as a lot of work was already done there.

cypress/e2e/Labels/tests/multipleLabelsCRUDTests.ts
susannahmarcus: Ah, it does clean them in the afterEach, which runs through and deletes all the labels (line 43)

app/ui/labels.py
susannahmarcus: I based the ordering off of the old ui in the native app. The order there was:
```
label name
group name -- rank
condition
```
that said, I'm open to suggestions.

joshelser (approved):
Python changes look fine. Thanks!

PR #331, opened 2023-09-20
Replace label SQL procedures with python procedures

joshelser:
- [x] two test failures to fix

ref #234


rymurr:
lgtm so far!

joshelser:
Restored the `PredefinedLabels` class and subjected them to the same validation as we do for normal labels (instead of just validating the Condition like we previously did).

app/crud/labels.py
```diff
@@ -214,3 +224,22 @@ def verify_time_fields(cls, time: datetime.datetime) -> datetime.datetime:
...
+        :param session:
+        :return:
+        """
+        df = session.sql("select * from internal.predefined_labels").to_pandas()
+        # Lowercase all of the column names
+        df.rename(
+            columns={col_name: col_name.lower() for col_name in df.axes[1]},
+            inplace=True,
+        )
+        for row in df.to_dict(orient="records"):
```
joshelser, line 243: Will want to consolidate with changes from #339

vqmarkman (approved):
Looks good to me. I was a bit skeptical about this change, but my feeling now is that it will be much easier to manage.

app/crud/labels.py
```diff
@@ -214,3 +224,22 @@ def verify_time_fields(cls, time: datetime.datetime) -> datetime.datetime:
     def enabled_or_dynamic(cls, b: bool) -> bool:
         assert isinstance(b, bool)
         return b
+
+
+class PredefinedLabel(Label):
+    @classmethod
+    def validate_all(cls, session: snowflake.snowpark.Session):
```
rymurr, line 231: can you create a ticket to move the predefined label def'n and creation out of sql and into this?

bootstrap/004_labels.sql
```diff
@@ -263,114 +166,66 @@ END;
...
+from crud import delete_entity
+def delete_label(session, name):
+    return delete_entity(session, 'LABEL', name)
+$$;
 
+-- TODO Differentiate from ungrouped labels  by having "AND IS_DYNAMIC" in the delete clause
 CREATE OR REPLACE PROCEDURE ADMIN.DELETE_DYNAMIC_LABEL(name text)
     RETURNS text
-    LANGUAGE SQL
+    language python
```
rymurr, line 185: why have this? Its the same is it not?

bootstrap/004_labels.sql
```diff
@@ -263,114 +166,66 @@ END;
...
-    CALL INTERNAL.UPDATE_LABEL_VIEW();
-    return 'done';
-END;
+$$
+from crud import delete_entity
+def delete_label(session, name):
+    return delete_entity(session, 'LABEL', name)
+$$;
 
 CREATE OR REPLACE PROCEDURE ADMIN.UPDATE_LABEL(oldname text, name text, grp text, rank number, condition text, is_dynamic boolean)
```
rymurr, line 198: is it possible to get rid of this one? Do we need the is_dynamic still?

rymurr (commented):
looks great, I think we can make a few more simplifcations otherwise :ship: it

bootstrap/004_labels.sql
joshelser: I had consolidated the `DELETE_DYNAMIC_LABEL` and `DELETE_LABEL` last weekend, and then regretted doing it in the first pass. I left them separate for this PR, but completely agree that we don't need two long-term.

bootstrap/004_labels.sql
joshelser: Same with this one -- intentionally didn't collapse the overloaded functions. If it's good with you, would do that in a follow-on (likely today)

app/crud/labels.py
joshelser: Of course

bootstrap/004_labels.sql
rymurr: foolow on is perfect

bootstrap/004_labels.sql
rymurr: foolow on is perfect

PR #338, opened 2023-09-20
Add default timezone for warehouse operations

### LINKED ISSUE #332: Warehouse Management - add timezone to config section
built off #29

rymurr:
closes #332

### DIFF
--- a/app/ui/pages/10_Settings.py
+++ b/app/ui/pages/10_Settings.py
@@ -3,6 +3,7 @@ import connection
 import config
 import sthelp
 import setup
+import pytz
 
 
 sthelp.chrome("Settings")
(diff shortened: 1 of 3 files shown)


app/ui/pages/10_Settings.py
```diff
@@ -82,6 +86,14 @@ def task_listing(
             key="serverless_credit_cost",
         )
         storage = st.text_input("Storage Cost (/tb)", value=tbcost, key="storage_cost")
+        timezone = st.selectbox(
+            "Default Timezone",
+            index=next(
+                i for i, v in enumerate(pytz.all_timezones) if v == timezone_start
```
joshelser, line 92: nit: Was glancing at `all_timezones` and see that it has over 500 entries in it. Would `common_timezones` be abrasive?

joshelser (approved):
I'm assuming that the timezone names that `pytz` returns are usable by Snowflake timezone types/functions, or maybe you're planning to do the core algorithm solely in Python and it doesn't matter if Snowflake's timezones align?

rymurr:
> I'm assuming that the timezone names that `pytz` returns are usable by Snowflake timezone types/functions, or maybe you're planning to do the core algorithm solely in Python and it doesn't matter if Snowflake's timezones align?

Both use the standard notation fo rtimezones

app/ui/pages/10_Settings.py
rymurr: fixed to use common_timezones. Makes a lot more sense

joshelser (approved):

PR #339, opened 2023-09-20
Add warehouse scheduling widget

rymurr:
Still needs more testing


rymurr:
Have added some tests, there are still some bugs that should be addressed in #349 #350 #351 #352 #353

app/ui/warehouses.py
```diff
@@ -0,0 +1,275 @@
...
+        warehouses[i] = args[0]
+        warehouses[i]._dirty = True
+        comment, new_warehouses = verify_and_clean(warehouses)
+        if comment is not None:
+            return comment
+        [i.update(connection.Connection.get(), i) for i in new_warehouses]
+
+    def list(self):
+        st.button(
+            "Clear State (TODO, for testing only)",
```
joshelser, line 213: need to make sure we drop this button before releasing.

joshelser (approved):
Overall, looks good. The biggest piece of feedback (centralizing validation of a collection of WarehouseSchedules) was something Ryan and I talking IRL and we both think this is a good idea.

I think it'll be easier to merge this and do cleanup as follow-on. I can help out here.

Issue #39, opened 2023-07-12
Unable to create ungrouped label in Cypress test


joshelser:
Have noticed Cypress tests that create an ungrouped label have failed with the error "Invalid SQL condition ... syntax error on line 3 at position 1 unexpected 'then'".

This has only been observed in Cypress. We're unsure if there is some workaround that can be done in the Cypress test to work around this or if it's some other kind of bug.

vqmarkman:
Assigned to myself to check in test and validate this behavior again.

vqmarkman:
The problem here is that (a) label creation worked when submitted manually (b) when ran with cypress automation, condition field ended up being empty despite the fact that input was typed in, sending empty string parameter to create_label stored procedure:
`CALL ADMIN.CREATE_LABEL('label_150c147d-9f36-48fd-8cfa-e2c6c5f57a22', NULL, NULL, '');`

Adding wait did not help, however, I noticed that st.text_area requires input submission, see below in blue:

![New label](https://github.com/sundeck-io/OpsCenter/assets/104043363/7508c1b9-e147-46c7-8746-2a6e7cf9c1e8)

This cypress workaround seems to work:

```
  cy.get('textarea[aria-label="Condition"]')
    .clear()
    .type(condition).type('{command+enter}'); <--- note sending "command+enter" after submission.
```

I will be adding test for ungrouped labels back shortly. cc @doronrosenberg

joshelser:
Great find, Vicky!

PR #347, opened 2023-09-21
299: Dechunk and fix probes e2e tests

### LINKED ISSUE #80: UI tests: Add validation of successful object creation in the UI
We need to add validation that objects have correct properties. 
For example, after group label has been created we need to make sure that group, rank and condition are exactly the same as we specified in the input. Currently tests only validate label name by deleting label from the page.

In addition to that, `checkGroupLabelNotExist(groupName)` is not doing the right thing. Fix it at the same time when validation is added.

susannahmarcus:
- dechunks the probes tests into smaller segments
- takes some of the labels utils/constants and moves them to the support file since they work in other scopes (when having to do with the listing pages, headers, and buttons)
- Adds ability to verify field values on the probes form (labels form already addressed by #328)
- Adds update tests to probes

closes #80

### DIFF
--- a/cypress/e2e/Labels/utilsAndConstants/labelsFormUtils.ts
+++ b/cypress/e2e/Labels/utilsAndConstants/labelsFormUtils.ts
@@ -1,10 +1,10 @@
-import { LABEL_FORM_FIELDS } from "./labelTestConstants";
+import { LABEL_FORM_FIELDS, UNGROUPED } from "./labelTestConstants";
 
 export const checkPresenceOfGroupNameInput = (options: {
   isPresent: boolean;
 }) => {
   const { isPresent = true } = options;
-  cy.dataId({ value: "stMarkdownContainer" })
+  cy.dataId("stMarkdownContainer")
     .contains(LABEL_FORM_FIELDS.GROUP_NAME)
     .should(($el) => {
       if (isPresent) {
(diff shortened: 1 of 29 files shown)


cypress/e2e/Labels/utilsAndConstants/labelsFormUtils.ts
```diff
@@ -14,3 +14,70 @@ export const checkPresenceOfGroupNameInput = (options: {
...
+    groupName !== UNGROUPED &&
+    cy.get('input[aria-label="Group Name"]').clear().type(groupName);
+
+  rank &&
+    cy.get('input[aria-label="Group Rank"]').clear().type(rank).type("{enter}");
+
+  cy.get('textarea[aria-label="Condition"]')
+    .clear()
+    .type(condition)
+    .type("{command+enter}");
```
doronrosenberg, line 38: will this work on non-mac?

cypress/e2e/Labels/utilsAndConstants/labelsFormUtils.ts
```diff
@@ -14,3 +14,70 @@ export const checkPresenceOfGroupNameInput = (options: {
...
+      .clear()
+      .type(newRank)
+      .type("{enter}");
+
+  newCondition &&
+    cy
+      .get('textarea[aria-label="Condition"]')
+      .clear()
+      .type(newCondition)
+      .type("{command+enter}");
```
doronrosenberg, line 67: ditto

cypress/e2e/Labels/utilsAndConstants/labelsFormUtils.ts
vqmarkman: @doronrosenberg , if you are referring to `command+enter` then yes, it works on mac and in GHA.
This was the fix for https://github.com/sundeck-io/OpsCenter/issues/39 (all the details on implementation are there)

cypress/e2e/Labels/utilsAndConstants/labelsUtils.ts
```diff
@@ -180,28 +137,19 @@ export function checkUpdatedLabelExists(options: {
...
-            .contains(condition)
-            .should("exist");
+          cy.dataId("stVerticalBlock").contains(condition).should("exist");
 
-        rank &&
-          cy
-            .dataId({ value: "stVerticalBlock" })
-            .contains(rank)
-            .should("exist");
+        rank && cy.dataId("stVerticalBlock").contains(rank).should("exist");
```
vqmarkman, line 149: @susannahmarcus , I don't understand what happened with formatting (and there are bunch of other cases like this).  Code in lines 195 - 199 is new since yesterday. 
Did you manually remove `{ value:` part and then because line became shorter it got collapsed into a single line? 
If you did not manually remove it, it's almost like another version `prettier` is doing it...

cypress/support/customCypressUtils.ts
```diff
@@ -7,32 +7,27 @@ declare global {
...
        */
-      dataCy(options: {
-        value: string;
-        timeout?: number;
-      }): Chainable<JQuery<HTMLElement>>;
-      dataId(options: {
-        value: string;
-        timeout?: number;
-      }): Chainable<JQuery<HTMLElement>>;
+      dataCy(value: string, timeout?: number): Chainable<JQuery<HTMLElement>>;
```
vqmarkman, line 10: another case: 
`options: {` <--- removed and reformatted
        value: string;
        timeout?: number;
})

vqmarkman (approved):
Awesome, ship it!

cypress/e2e/Labels/utilsAndConstants/labelsUtils.ts
susannahmarcus: formatting:
in JS, if you're looking to see if something exists, you can just have the check structured like this. It basically says `if rank is defined, do this`.
As for the `{ value: "stVerticalBlock" }` turning into `"stVerticalBlock" ` it's for the custom utils I've added. I changed their arguments structure because I made it an object before since there was a optional timeout increase, but noticed that nothing was using it and it was adding the step of typing `{value:}` for nothing.

cypress/support/customCypressUtils.ts
susannahmarcus: Yeah, I did this manually, as specified in the last comment

doronrosenberg (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "devDependencies": {
-    "cypress": "^12.16.0",
+    "cypress": "^13.2.0",
```
vqmarkman, line 3: @susannahmarcus , you might want to enable videos again. In this version they are turned off by default. Only screenshots are recorded.

```video: true```

vqmarkman: never mind, I see it.

vqmarkman (approved):

PR #348, opened 2023-09-22
Dont run proes if underlying infra isnt ready

rymurr:


rymurr:
Merging to do a patch release. Happy to address comments after

joshelser:
Good find 👍🏼

PR #354, opened 2023-09-22
Warehouse Management Task

### LINKED ISSUE #336: Warehouse management - create task to actually manage warehouse settings
continuing from #29

rymurr:
closes #336


app/crud/wh_sched.py
```diff
@@ -0,0 +1,314 @@
...
+    return to_run.reset_index()
+
+
+def build_ts(today, now, x):
+    if x < now.time():
+        return datetime.datetime.combine(today, x).replace(tzinfo=now.tzinfo)
+    else:
+        return datetime.datetime.combine(today - datetime.timedelta(days=1), x).replace(
+            tzinfo=now.tzinfo
+        )
```
joshelser, line 228: I've been going over this method, and I can't seem to figure out what exactly you were trying to do in the else-branch.

In the if-branch, for each schedule in a warehouse, you're fixing the start time to today's date and then switching the timezone. Makes sense.

In the else branch, you're the schedule(s) which have not _yet_ run to yesterday's date and then switching the timezone.

My interpretation would be that we want to always do `datetime.datetime.combine(today, x).replace(tzinfo=now.tzinfo)` and remove the `if/else` branch. The "should_run" column in the next line in the calling function would naturally filter the schedules which should not _yet_ run. What am I missing? :)

PR #360, opened 2023-09-22
Query hash again

joshelser:
rel to #323

### DIFF
--- a/bootstrap/042_utils.sql
+++ b/bootstrap/042_utils.sql
@@ -22,18 +22,15 @@ as
 ;
     end;
     $$;
-else
-    execute immediate $$
-    begin
-create or replace function tools.is_reoccurring_query(qph varchar, size number)
-returns boolean
-as
-'false';
-create or replace function tools.is_ad_hoc_query(qph varchar, size number)
-returns boolean
-as
-'false';
-    end;
-    $$;
-end if;
+    end if;
 end;
+
+create function if not exists tools.is_reoccurring_query(qph varchar, size number)
+returns boolean
+as
+'false';
+
+create function if not exists tools.is_ad_hoc_query(qph varchar, size number)
+returns boolean
+as
+'false';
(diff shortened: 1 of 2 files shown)


joshelser:
It seems like this rabbit hole goes deep yet again.

This patch fixes the first issue, but I'm seeing that the ordinal position between internal_reporting and internal_reporting_mv does _not_.

```
-- returns 78 to 81
call admin.run_as_app($$ SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = 'INTERNAL_REPORTING' AND TABLE_NAME = 'QUERY_HISTORY_COMPLETE_AND_DAILY' and column_name ilike '%hash%' order by ordinal_position $$);
-- returns 82 to 85
call admin.run_as_app($$ SELECT COLUMN_NAME, DATA_TYPE, ORDINAL_POSITION FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = 'INTERNAL_REPORTING_MV' AND TABLE_NAME = 'QUERY_HISTORY_COMPLETE_AND_DAILY'  and column_name ilike '%hash%' order by ordinal_position $$);
```

I think this is a separate problem.

joshelser:
on re-create the ordinal position now match up again, going to set this one aside as a different problem.

joshelser:
Testing:
- [ ] install v1.24 with materialization, upgrade to this patch, run commands to enable the bundle and finalize setup, observe success.
- [ ] install v1.24 with materialization,  commands to enable the bundle and finalize setup, observe failure, upgrade to this patch, observe success.
- [ ] fresh install of patch with bundle enabled, observe success
- [ ] fresh install of patch with bundle disabled, run commands to enable, observe success

rymurr (approved):

PR #370, opened 2023-09-26
Don't include deleted users, they may be duplicates

### LINKED ISSUE #368: Not getting email notification from the probe when OpsCenter is connected to Sundeck tenant via Snowflake SSO
Probe definition:
Screenshot 2023-09-25 at 5 05 08 PM

Result of ```call admin.run_as_app('select * from internal.probe_actions');```

Screenshot 2023-09-25 at 5 04 25 PM

Query is cancelled:
```
test#COMPUTE_WH@(no database).(no schema)>select /* vicky - email notification only  */ system$wait(180);
000604 (57014): SQL execution canceled                                          
```

email is not sent.

If I don't select check box to cancel the query, email is not sent and there is no record in ```internal.probe_actions``` table.

* Same works for me when OpsCenter is connected to a **new**  Sundeck tenant created with email/password.
* SSO tenant: **d3gyl69h** (us-west-2) on prod, created on 08/08/2023

rymurr:
closes #368

### DIFF
--- a/bootstrap/027_user_emails.sql
+++ b/bootstrap/027_user_emails.sql
@@ -5,7 +5,7 @@ CREATE OR REPLACE PROCEDURE internal.refresh_users() RETURNS STRING LANGUAGE SQL
 BEGIN
     BEGIN TRANSACTION;
         truncate table internal.sfusers;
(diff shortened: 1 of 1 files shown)


vqmarkman (approved):
wow ... ship it!

PR #374, opened 2023-09-26
Warehouse Schedule Task Testing

joshelser:
* Python function which embodies the task itself
* Testable via dataframe mocked data

- [x] Need to flesh out the unit tests

Edit: this ended up getting really big after writing a first pass of tests. I don't want to layer more functionality into this patch.

Overtakes #354
Ref #336


app/crud/wh_sched.py
```diff
@@ -130,25 +133,222 @@ def verify_scales(cls, values):
...
+
+
+def update_warehouse(warehouse_now, warehouse_next):
+    set_stmts = compare_warehouses(warehouse_now, warehouse_next)
+    if set_stmts:
+        return f"alter warehouse {warehouse_next.name} set {set_stmts}"
+    return ""
+
+
+class WarehouseSchedulesTask:
```
rymurr, line 205: I am confused about why this class has to exist. It has no state at all

rymurr (commented):
lgtm, 1 comment.

I assume conditionally enabling the task will be a second PR?

app/crud/wh_sched.py
joshelser: Sorry, I forgot to call this out when publishing the PR. I was struggling to get the mocking to work with the logic as module functions (not sure if that's correct terminology -- as your original draft PR had implemented). I had already wasted a couple of hours fiddling with mocking instead of writing tests. Making this a class made the mocking much easier to set up. I would not normally advocate for product changes just to enable testing, but this felt like an appropriate exception given the time I had already spent.

joshelser:
> I assume conditionally enabling the task will be a second PR?

Yes

app/crud/wh_sched.py
rymurr: Annoying, but I agree. Mind just putting a comment to that effect in the class pydoc. We can fix it later if we are so inclined

joshelser: of course.

rymurr (approved):

PR #380, opened 2023-09-27
321: Updated opscenter ReadMe for addition of typescript

### LINKED ISSUE #321: Update README file for Cypress - typescript is a requirement now
I checked out main and got this error:

```
Error: You are attempting to run a TypeScript file, but do not have TypeScript installed. Ensure you have 'typescript' installed to enable TypeScript support.
```
To install:

`npm install typescript --save-dev`

susannahmarcus:
closes #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,11 @@ dbname = "..."
 ```
 
 Then run the following commands:
+
 ```
+# Get node modules for the UI
+npm install
+
 # Create dev database
 poetry run python deploy/devdeploy.py -p local_dev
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,7 +69,11 @@ dbname = "..."
 ```
 
 Then run the following commands:
+
 ```
+# Get node modules for the UI
+npm install
```
joshelser, line 75: This is only necessary if you want to run Cypress tests, right?

What about adding a `#### Running Cypress Tests` section  beneath `### Local Development outside of App` which briefly explains that you can run tests using Cypress which exercise OpsCenter and you can run them, e.g.
```
npm install # first time only
yarn cypress run
```

(or whatever the expected incantation is)

susannahmarcus: good point. I put it there since they have a shared package.json, but this is also already covered in the cypress readme. do I need to add it here as well?

joshelser: I didn't realize there was a `cypress/README.md`. Completely agree with you, adding this info here is unnecessary. I don't think we need to do anything...

PR #381, opened 2023-09-27
Enable warehouse scheduling task

joshelser:
Need to write more tests and verify the queries actually work. Just posting as draft if you have any feedback.

### DIFF
--- a/app/ui/warehouses.py
+++ b/app/ui/warehouses.py
@@ -1,9 +1,9 @@
+import os
 import uuid
 import streamlit as st
 import connection
 from crud.wh_sched import (
     WarehouseSchedules,
-    WarehouseSchedulesTask,
     _WAREHOUSE_SIZE_OPTIONS,
 )
 from table import build_table, Actions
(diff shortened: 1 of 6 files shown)


joshelser:
I think this is still ready for review. It doesnt' work end-to-end (as I hoped). The scheduling logic in python should be able to stay in python.

app/ui/warehouses.py
```diff
@@ -214,14 +227,23 @@ def on_update_click_internal(self, *args) -> Optional[str]:
...
+            update_task_state(conn)
 
     def list(self):
-        st.button(
-            "Clear State (TODO, for testing only)",
-            on_click=lambda: connection.execute(
-                f"delete from internal.{WarehouseSchedules.table_name}"
-            ),
-        )
+        if os.environ.get("OPSCENTER_LOCAL_DEV", False):
```
rymurr, line 234: I thikn we should still remove this before publishing.

bootstrap/090_post_setup.sql
```diff
@@ -304,6 +304,40 @@ exception
...
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "X-Small"
+    AS
+    execute task TASKS.WAREHOUSE_SCHEDULING;
+
+-- The real warehouse scheduling logic cannot be in python with serverless task. Make a dummy task body.
+CREATE OR REPLACE TASK TASKS.WAREHOUSE_SCHEDULING
+    ALLOW_OVERLAPPING_EXECUTION = FALSE
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "X-Small"
+    as
+    select 1;
```
rymurr, line 339: I assume this will be filled in later?

app/ui/warehouses.py
joshelser: Agree. As Robert has started showing stuff off in "pre-release", I just wanted to consolidate some of my local hacks to hide this. Will go away in the end for sure.

bootstrap/090_post_setup.sql
joshelser: Right, this will be changed to call the (pending) SQL procedure which mimics what the python procedure does (since we can't call a python proc from a serverless task).

I realized that the scheduling of these tasks (which _can_ be in python) should be reviewed on its own, rather than trying to push everything in at the same time.

joshelser:
rebased to get the cypress changes

rymurr (approved):

PR #386, opened 2023-09-28
Rewrite the python warehouse schedule task body in SQL

joshelser:
Not nearly complete. Only does the selection of schedules to consider, none of the building up of `alter warehouse` commands.

On top of #381

### DIFF
--- a/bootstrap/011_warehouse_schedules.sql
+++ b/bootstrap/011_warehouse_schedules.sql
@@ -3,19 +3,6 @@
 
 -- CREATE TABLE IF NOT EXISTS internal.task_warehouse_schedule(RUN TIMESTAMP_LTZ(9), SUCCESS BOOLEAN, OUTPUT VARIANT)
 
-CREATE OR REPLACE PROCEDURE INTERNAL.UPDATE_WAREHOUSE_SCHEDULES()
-    RETURNS VARIANT
-    language sql
- AS
-DECLARE
-    task_outcome variant;
-    this_run timestamp_ltz default (select current_timestamp());
-    last_run timestamp_ltz default (select run from internal.task_warehouse_schedule order by run desc limit 1);
-BEGIN
-    call internal.UPDATE_WAREHOUSE_SCHEDULES(:last_run, :this_run) into :task_outcome;
-    return task_outcome;
-END;
-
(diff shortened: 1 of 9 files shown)


joshelser:
thinking aloud:
* I want to be able to pass in a `timestamp` from outside, else writing tests to cover these procedures is nary impossible.
* Passing in a `TIMESTAMP_LTZ` (which is what is defined in the task table) appears to result in mangled dates where the original wall clock time is converted to UTC wall clock time, but the timezone is left as is (e.g. if default timezone is -7, the wall clock gets 7hours added to it but timezone still claims -7).
* The warehouse schedules are defined without timezones, but need to be shifted with the opscenter defined timezone.
* When the task runs, it needs to pick the appropriate schedule(s) based off of the last time the task ran and "now".
* Timestamp_ltz's appear to be reasonable _within_ the stored procedure. So maybe this is just an issue with injecting timestamps into the procedure for testing?

joshelser:
In a nutshell:

* Every time the warehouse schedules are changed, we generate `ALTER WAREHOUSE` statements for each schedule in the system and store them in a table, referenced back to the `wh_schedules` table with the same `id_val` column.
* SQL proc takes the current time and the last time the task ran, and sees if there are any schedules which should run now.
* For every schedule found, we `execute immediate ..` the `alter warehouse` statement and record an outcome into the task_warehouse_schedule table.
* A very simple unit test (that is a bit contrived as we're lacking the admin procedures for CRUD of warehouse schedules)

bootstrap/011_warehouse_schedules.sql
```diff
@@ -1,18 +1,96 @@
...
-$$
-from crud import WarehouseSchedulesTask
-def run(session):
-    task = WarehouseSchedulesTask(session)
-    return task.run()
-$$;
+DECLARE
+    task_outcome variant default (select object_construct());
+BEGIN
+    -- The task calls this procedure with NULL and lets the procedure figure out the details.
```
rymurr, line 13: i wonder if this might be neater using a session parameter?

rymurr (approved):
one minor comment hat can be addressed later

bootstrap/011_warehouse_schedules.sql
joshelser: Rather than passing in `NULL`, you mean? I didn't think we could do that in an owner's rights procedure.

rymurr: that null is hella ugly so we have to think of something to do.

PR #390, opened 2023-09-29
Persist current 'enabled' value when creating new schedules

joshelser:
We were not including the `enabled` attribute from the UI when creating a new schedule.

Then, potentially had some schedules which were enabled and one that was disabled. Flipping the checkbox would then invert all of the schedules and leave you with one enabled and the rest disabled. Changed the logic to read the checkbox's state instead of just inverting the enabled value in sql.

Also fixes a simple bug when switching to a different warehouse (generated a stacktrace) which was because we tried to get a connection while already having a connection.


rymurr (approved):

joshelser:
this still throws the error

> Verify failed: 'st.session_state has no key "enabled". Did you forget to initialize it? More info: https://docs.streamlit.io/library/advanced-features/session-state#initialization'

apparently I need somethign fancier.

rymurr (approved):

PR #394, opened 2023-09-30
Warehouse Scheduling phase 1.5

rymurr:
Create new filters section and put heatmap/schedule into tabs.

Also refactored tests to not require the class


app/crud/wh_sched.py
```diff
@@ -144,6 +144,7 @@ def create_table(cls, session, with_catalog_views=True):
 
 
 def describe_warehouse(session: Session, warehouse: str):
+    print(f"Describing warehouse {warehouse}")
```
joshelser, line 147: nit: remove

app/crud/wh_sched.py
```diff
@@ -286,155 +285,143 @@ def _make_cron_schedule(
...
-        statements = [
-            f"alter task if exists {cls.task_name}_{offset} suspend;"
-            for offset in cls.task_offsets
-        ]
-        statements.insert(0, "begin")
-        statements.append("end;")
-        session.sql("\n".join(statements)).collect()
+
+task_name: ClassVar[str] = "TASKS.WAREHOUSE_SCHEDULING"
+task_offsets: ClassVar[List[int]] = [0, 15, 30, 45]
```
joshelser, line 308: Should these just be `str` and `List[int]` now that the `class` is removed?

app/ui/warehouse_schedule.py
```diff
@@ -0,0 +1,131 @@
...
+        count(distinct size) as sz_count
+    from internal.wh_schedules group by warehouse_name
+     ), wh_agg as (
+        select warehouse_name, enabled, autoscale, case when sz_count=1 then size else 'variable'end as size from wh_raw
+     )
+    select * from all_wh
+     left join util on util.warehouse_name = all_wh.warehouse_name
+      left join wh_agg on all_wh.warehouse_name = wh_agg.warehouse_name;
+      let x resultset := (select warehouse_name as warehouse,
+            coalesce(size, "size") as size,
```
joshelser, line 93: What does it mean if we don't get a warehouse size? Returning `"size"` seems prone to error as nothing downstream would expect this value.

app/ui/warehouse_schedule.py
```diff
@@ -0,0 +1,131 @@
...
+            cols[2].button(
+                "90", on_click=sess.set_date_range_days, args=[90], key="last90"
+            )
+            cols[3].button(
+                "365", on_click=sess.set_date_range_days, args=[365], key="last365"
+            )
+
+        credit_cost = config.get_compute_credit_cost()
+        sql = f"""
+        with tmp as procedure()
```
joshelser, line 66: This is pretty hideous, but I see why you're doing it this way. I also had no idea this syntax existed.

I'm worried about this being dubious to test/maintain.

app/ui/warehouse_schedule.py
```diff
@@ -0,0 +1,131 @@
...
+            coalesce(enabled, false) as enabled
+        from table(result_scan(last_query_id())));
+      return table(x);
+    end;$$ call tmp();
+     """
+
+        sql = connection.Connection.bind(sql, {"start": range[0], "end": range[1]})
+        df = connection.execute_with_cache(sql)
+        df.columns = [c.lower() for c in df.columns]
+        data = [WarehouseSummary.parse_obj(dict(**i)) for i in df.to_dict("records")]
```
joshelser, line 106: We should push this pattern down into the BaseModel. We have it in a couple of places already.

joshelser (commented):
Sorry, my PRs created a bunch of merge conflicts here.

WDYT about that big inline procedure. Is that something worth separating into SQL that we could try to add test coverage for?

app/ui/warehouse_schedule.py
rymurr: this is probably worth explaining, I will clean it up. size is the size according to our table. If our table doesn't have an entry we use "size" which comes from the CTE on line 73 which is its actual current size. The `show` commands all return lowercase columns for whatever reason. I will rename it in the CTE

rymurr: lmk if its ok now

app/ui/warehouse_schedule.py
rymurr: what is hideous? Not sure I follow

app/ui/warehouse_schedule.py
rymurr: good idea, fixed

app/crud/wh_sched.py
rymurr: yup, fixed

rymurr:
I've been wondering about the big inline SQL in general (not just the anonymous proc). I've been thinking we should prob centralise them somehow. Either in a new schema (then the anonymous proc becomes a UDTF) or in a new set of files somewhere. Was waiting for more signal before doing so though.

app/ui/warehouse_schedule.py
joshelser: My current line of thinking is that we want to stop from introducing "heavy" logic into OpsCenter code because Streamlit is kind of a poison pill when it comes to dev-friendly testing.

I shouldn't have used the word "hideous" -- I meant it as "very complex" and that complexity would be something I'd like to be able to run some kind of test case over. However, now that I'm reading this again: this would be pretty gross to try to write a test to cover without putting together some fixture for warehouse sessions.

The other thing unclear to me was why you used an inline procedure and not a scripting block. Is that an API limitation?

joshelser (dismissed):
Thanks for rebasing this and introducing that `from_df(DataFrame)` helper method.

On second read, I'm tempering my reaction to the big, inline procedure because I don't see something better that could be done instead.

app/ui/warehouse_schedule.py
rymurr: ive removed the inline procedure in favour of a  scripting block

app/ui/warehouse_schedule.py
joshelser: oh duh, I read that original line as if it were `coalesce(size, 'size')`. Thanks, yes, it's clearer now.

joshelser (approved):

joshelser:
> I've been wondering about the big inline SQL in general (not just the anonymous proc). I've been thinking we should prob centralise them somehow. Either in a new schema (then the anonymous proc becomes a UDTF) or in a new set of files somewhere.

Yeah, I think your reaction on what to do and when to do it are both correct, but I'll continue to ask (else I think we may miss the inflection point) :)

PR #397, opened 2023-10-02
Treat warehouse schedules as in the opscenter-configured timezone.

### LINKED ISSUE #396: Run warehouse schedules per the configured timezone in OpsCenter
Part of warehouse scheduling -- we introduced a timezone configuration into OpsCenter. The times in the warehouse schedules should be applied with respect to that timezone (not the default timezone of the Snowflake Account).

joshelser:
1. All warehouse schedule tasks should be marked as running in the configured opscenter timezone.
2. When the task runs, convert the current time and the last task run time to the configured timezone. Then, apply those times against the task schedule and execute the logic per usual.

Adds some (contrived) SQL unit tests to validate the logic of this task before we have the admin procedures.

Closes #396

### DIFF
--- a/app/crud/wh_sched.py
+++ b/app/crud/wh_sched.py
@@ -3,6 +3,8 @@ from typing import ClassVar, List, Optional
 import datetime
 from .base import BaseOpsCenterModel
 from pydantic import validator, root_validator, Field
+from pytz import timezone
+from pytz.exceptions import UnknownTimeZoneError
 from snowflake.snowpark.functions import col, max as sp_max
 from snowflake.snowpark import Row, Session
 import pandas as pd
@@ -319,12 +321,30 @@ def regenerate_alter_statements(session: Session, schedules: List[WarehouseSched
     WarehouseAlterStatements.batch_write(session, alter_stmts, overwrite=True)
 
 
-def update_task_state(session: Session, schedules: List[WarehouseSchedules]) -> bool:
+def get_schedule_timezone(session: Session) -> timezone:
+    """
+    Fetch the 'default_timezone' from the internal.config table. If there is no timezone set or the
+    timezone which is set fails to parse, this function will return the timezone for 'America/Los_Angeles'.
+    """
+    str_tz = session.call("get_config", "default_timezone") or "America/Los_Angeles"
+    try:
+        return timezone(str_tz)
+    except UnknownTimeZoneError:
+        return timezone("America/Los_Angeles")
+
+
+def update_task_state(
+    session: Session, schedules: List[WarehouseSchedules], tz=None
+) -> bool:
     # Make sure we have at least one enabled schedule.
     enabled_schedules = [sch for sch in schedules if sch.enabled]
     if len(enabled_schedules) == 0:
         disable_all_tasks(session)
         return False
+
+    # Indirection for unit tests. Caller is not expected to provide a timezone.
+    if not tz:
+        tz = get_schedule_timezone(session)
 
     # Build the cron list for the enabled schedules
     alter_statements = []
@@ -333,7 +353,7 @@ def update_task_state(session: Session, schedules: List[WarehouseSchedules]) -> 
         #   1. ALTER TASK ... SUSPEND
         #   2. ALTER TASK ... SET SCHEDULE = 'USING CRON ...'
         #   3. ALTER TASK ... RESUME
-        alter_statements.extend(_make_alter_task_statements(schedules, offset))
+        alter_statements.extend(_make_alter_task_statements(schedules, offset, tz))
 
     # Collect the statements together
     alter_body = "\n".join(alter_statements)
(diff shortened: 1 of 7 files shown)


app/crud/wh_sched.py
```diff
@@ -319,21 +321,39 @@ def regenerate_alter_statements(session: Session, schedules: List[WarehouseSched
...
+def get_schedule_timezone(session: Session) -> timezone:
+    """
+    Fetch the 'default_timezone' from the internal.config table. If there is no timezone set or the
+    timezone which is set fails to parse, this function will return the timezone for 'America/Los_Angeles'.
+    """
+    str_tz = session.call("get_config", "default_timezone") or "America/Los_Angeles"
+    try:
+        return timezone(str_tz)
+    except UnknownTimeZoneError:
+        return timezone("America/Los_Angeles")
```
rymurr, line 333: we can't know the account timezone?

bootstrap/011_warehouse_schedules.sql
```diff
@@ -10,23 +10,36 @@ CREATE OR REPLACE PROCEDURE INTERNAL.UPDATE_WAREHOUSE_SCHEDULES(last_run timesta
 DECLARE
     task_outcome variant default (select object_construct());
 BEGIN
+    -- Get the configured timezone or default to 'America/Los_Angeles'
+    let tz text;
+    call internal.get_config('default_timezone') into :tz;
+    if (tz is null) then
+        tz := 'America/Los_Angeles';
+    end if;
+    task_outcome := (select object_insert(:task_outcome, 'opscenter timezone', :tz));
```
rymurr, line 19: I am not sure why we have two lines here. as in they are basically the same info right?

bootstrap/011_warehouse_schedules.sql
```diff
@@ -94,3 +107,12 @@ EXCEPTION
...
         RAISE;
 END;
+
+-- owners rights procedures can't get the timezone from the session parameters.
+CREATE OR REPLACE FUNCTION INTERNAL.GET_CURRENT_TIMEZONE()
+RETURNS STRING
+LANGUAGE JAVASCRIPT
+AS
+$$
+    return Intl.DateTimeFormat().resolvedOptions().timeZone;
```
rymurr, line 117: what does this return?

bootstrap/011_warehouse_schedules.sql
```diff
@@ -10,23 +10,36 @@ CREATE OR REPLACE PROCEDURE INTERNAL.UPDATE_WAREHOUSE_SCHEDULES(last_run timesta
...
+        tz := 'America/Los_Angeles';
+    end if;
+    task_outcome := (select object_insert(:task_outcome, 'opscenter timezone', :tz));
+    task_outcome := (select object_insert(:task_outcome, 'account timezone', internal.get_current_timezone()));
+
     -- The task calls this procedure with NULL and lets the procedure figure out the details.
     -- The ability to specify timestamps is only to enable testing.
     if (this_run is NULL) then
-        this_run := (select current_timestamp());
+        this_run := (select CONVERT_TIMEZONE(internal.get_current_timezone(), :tz, current_timestamp()));
```
rymurr, line 25: whats the behaviour of this when we pass in an LTZ timezone instead of an NTZ?

rymurr (dismissed):
couple of minor comments. Can be addressed in a follow up

app/crud/wh_sched.py
joshelser: Ah, here we can probably figured it out from Snowflake. I hadn't thought about checking the snowflake default.

bootstrap/011_warehouse_schedules.sql
joshelser: Ah, one is recording what the configured timezone for OpsCenter is and the other is recording what the default timezone is for the user's Snowflake account. I thought knowing this might be helpful in debugging issues in the future.

bootstrap/011_warehouse_schedules.sql
joshelser: This returns a timezone identifier like "America/New_York" (or similar).

bootstrap/011_warehouse_schedules.sql
joshelser: For the 3-arg CONVERT_TIMEZONE function, I don't believe there is any difference if you pass an LTZ or an NTZ (because the timestamp is assumed to be in the timezone specified by the first argument).

I _think_ changing this procedure to accept an NTZ instead of an LTZ is the way we solve the goofiness of the timestamps shifting when given to a procedure. I had been messing around with this over the weekend which works as I expect:
```
create or replace procedure internal.test_tz(now timestamp_ntz)
returns table(tz_ts TIMESTAMP_NTZ, ts timestamp_ltz, tz_ts2 TIMESTAMP_NTZ, ts2 TIMESTAMP_NTZ)
language sql
as
begin
    let res resultset := (
        select convert_timezone(internal.get_Current_timezone(), 'America/New_York', current_timestamp()),
                current_timestamp(),
                convert_timezone(internal.get_current_timezone(), 'America/New_York', :now),
                :now);
    return table(res);
end;
```

joshelser:
Going to ask for forgiveness and go ahead with the merge without re-approval since I only made a change to the test teardown code.

I filed #398 as follow-on. Happy to file others if I've missed a request.

PR #401, opened 2023-10-03
Separately MERGE ungrouped/grouped and dynamic grouped predefined labels

### LINKED ISSUE #400: migrate_predefined_labels can create duplicate dynamic grouped labels
When running `finalize_setup()` when the homepage loads, I've noticed I will sometimes get `duplicate column name: 'dbt Models'` and there will be multiple rows for the predefined dynamic grouped labels.

I believe the MERGE condition is wrong as it is looking only at `name` which is `NULL` for all dynamic grouped labels (rather than consider name+group_name)

joshelser:
Because the MERGE only inspected the label name (which is null for predefined grouped labels), we were incorrectly falling into the not matched condition and inserting duplicate rows.

Closes #400

### DIFF
--- a/bootstrap/004_labels.sql
+++ b/bootstrap/004_labels.sql
@@ -338,8 +338,9 @@ BEGIN
         RETURN FALSE;
     END IF;
 
+    -- Merge in ungrouped and grouped predefined labels
     MERGE INTO internal.labels t
-    USING internal.predefined_labels s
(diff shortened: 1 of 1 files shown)


joshelser:
Draft until have test this a little more.

I think that the name vs. group_name uniqueness on internal.labels will continue to bite us, so I'm also reticent to band-aid this yet-again.

PR #407, opened 2023-10-03
165: make github yaml file changes needed to run newest e2e test changes

susannahmarcus:


joshelser (changes requested):
```
      - name: Cypress run
        uses: cypress-io/github-action@v6
        env:
          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
          SNOWFLAKE_USERNAME: ${{ secrets.SNOWFLAKE_USERNAME }}
          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
        with:
          start: yarn run cypress
          browser: chrome
          headless: true
          record: false
```

Looking at this again, can you please update it to be this instead? You can completed remove the `Set cypress env variables` step. (in both places)

I dont' want to rely on the detection and redaction of the snowflake password across steps in a workflow. If we can limit which steps receive the password, that's better.

joshelser (approved):
tyvm 🚢

PR #410, opened 2023-10-03
Delete latest schedule

joshelser:


joshelser:
7bcabd1 is the only one we care about. includes #404

joshelser:
rebased since #404 landed

rymurr (approved):

PR #413, opened 2023-10-04
Prevent multiple main checks from running concurrently

joshelser:
Have seen some test-deploy's fail but always pass on re-run. I think it's because the Main Build checks are running concurrently and the test-deploy is not using a unique name. Band-aid for the moment.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,5 +1,7 @@
 name: Main Build Check
 
+concurrency: main
+
 on:
   push:
     branches: [main]


vqmarkman:
Just to clarify my understanding: only one deployment will be running concurrently, correct?

vqmarkman (approved):

joshelser:
> Just to clarify my understanding: only one deployment will be running concurrently, correct?

Yes, my expectation is that we will run only on "Main" workflow at a time.

PR #414, opened 2023-10-04
Make sure the frontend properly handles create when verify_and_clean …

joshelser:
…does not adjust the new schedule

Generate a new id_val up front and use that instead of checking for the empty string.

Fixes the StopIteration we used to get if you tried to create a new schedule that didn't use Inherit
```
...
File "/Users/elserj/projects/opscenter.git/app/ui/base.py", line 113, in on_create_click
    outcome = self.on_create_click_internal(*args)
File "/Users/elserj/projects/opscenter.git/app/ui/warehouses.py", line 196, in on_create_click_internal
    new_row = next(i for i in new_data if i.id_val == "")
```

### DIFF
--- a/app/ui/warehouses.py
+++ b/app/ui/warehouses.py
@@ -175,9 +175,13 @@ class Warehouses(Container):
         return new_update
 
     def on_create_click_internal(self, *args) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


joshelser:
I want to get a release staged today, so I'm going to merge this without review. Please let me know if you have any feedback, thanks!

PR #419, opened 2023-10-04
Create a simple reporting view for the warehouse schedules task

joshelser:


joshelser:
Screenshot 2023-10-04 at 5 31 25 PM

example row from the view

vqmarkman (approved):

joshelser:
Thanks for the catch on variant type, Vicky. Fixed it so they have reasonable types now.

bootstrap/011_warehouse_schedules.sql
```diff
@@ -7,11 +7,16 @@ CREATE OR REPLACE PROCEDURE INTERNAL.CREATE_WAREHOUSE_SCHEDULES_VIEWS()
...
     CREATE OR REPLACE VIEW catalog.warehouse_schedules COPY GRANTS AS SELECT * exclude (id_val, day) FROM internal.wh_schedules;
+    CREATE OR REPLACE VIEW reporting.warehouse_schedules_task_history as
+        SELECT run, success, output:"statements"::ARRAY as statements_executed, output:"opscenter timezone"::TEXT as schedule_timezone, output:"warehouses_updated"::NUMBER as warehouses_updated
+        from internal.task_warehouse_schedule;
     -- Because we defer creation of the table until FINALIZE_SETUP, we need to re-run
     -- the grant commands to ensure that the user can see this view because it would
     -- not receive the grant during 100_final_perms.
     GRANT SELECT ON ALL VIEWS IN SCHEMA CATALOG TO APPLICATION ROLE ADMIN;
     GRANT SELECT ON ALL VIEWS IN SCHEMA CATALOG TO APPLICATION ROLE READ_ONLY;
+    GRANT SELECT ON ALL VIEWS IN SCHEMA REPORTING TO APPLICATION ROLE ADMIN;
```
rymurr, line 18: why don't we just use future grants in final perms?

joshelser: Because this procedure is being called from `admin.finalize_setup()` because the create tables are happening in Python (and `finalize_setup()` is the earliest we can invoke the Python code).

I'm of the mind that table+view definitions may be better served to live in SQL for that reason. If we move table/view defn's back to SQL, then we don't need the re-`GRANT`s.

rymurr: I am not sure I understand `GRANT SELECT ON FUTURE VIEWS IN SCHEMA...` is designed for the use case you describe

joshelser: Ah, yes, I agree with you if we add `ON FUTURE`. I'm not sure if there's a reason to not do that for catalog and reporting. Let me see.

PR #426, opened 2023-10-05
Add grants on future views and simplify WHSch perms.

### LINKED ISSUE #425: Switch to `grant select on future views` in final perm setup
Some tables/views are created when `admin.finalize_setup()` runs which occurs after 100_final_perms is called. However, the grants executed in 100_final_perms do not currently use the `ON FUTURE` option for the grant which means that the user-facing views created in `admin.finalize_setup()` are not automatically seen by the user.

Can't we simply do `grant select on future views` in 100_final_perms (and be careful to not add non-user-facing views to these two schemas)?

joshelser:
Closes #425

### DIFF
--- a/bootstrap/011_warehouse_schedules.sql
+++ b/bootstrap/011_warehouse_schedules.sql
@@ -6,17 +6,12 @@ CREATE OR REPLACE PROCEDURE INTERNAL.CREATE_WAREHOUSE_SCHEDULES_VIEWS()
     EXECUTE AS OWNER
 AS
 BEGIN
+    -- Create the catalog view.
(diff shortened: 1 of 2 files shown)


rymurr (approved):

joshelser:
Curious if you have any ideas here, @rymurr 

```
call admin.run_as_app($$ grant select on future views in schema REPORTING to APPLICATION ROLE ADMIN $$);
```

Fails with:
```
Insufficient privileges to operate on schema 'REPORTING'
```

I'm not finding any explanation in https://docs.snowflake.com/en/sql-reference/sql/grant-privilege-application-role that talks about limitations. My only guess is that maybe there's something on versioned schemas?

joshelser:
Seems like `grant * on future` doesn't work on versioned schemas, so I'm just going to close this for now. Doesn't seem like there's a simplification we can do right now.

PR #432, opened 2023-10-05
Refresh the warehouse schedule row when it hasn't been changed

### LINKED ISSUE #351: If warehouse is inactive and only has default rows, update the rows to current value of warehouse

joshelser:
If the user has only viewed the warehouse schedules, they will have a row persisted into internal.wh_schedules. However, we want to update that row with the latest state of the warehouse if the user has not enabled the schedule (implying they want it to take effect).

Closes #351

### DIFF
--- a/app/crud/wh_sched.py
+++ b/app/crud/wh_sched.py
@@ -306,17 +306,15 @@ def convert_time_str(time_str) -> datetime.time:
 
 def fetch_schedules_with_defaults(
     session: Session, warehouse: str
-) -> Tuple[List[WarehouseSchedules], bool]:
+) -> List[WarehouseSchedules]:
     """
-    Returns the WarehouseSchedules for the given warehouse and true if the schedules returned simply match the current
-    state of the warehouse.
+    Returns the WarehouseSchedules for the given warehouse.
     :param session: Snowpark session instance.
     :param warehouse: The snowflake warehouse to filter on.
     :return: A list of WarehouseSchedules for the given warehouse, creating and persisting
     default schedules if none exist.
     """
     schedules = WarehouseSchedules.find_all(session, warehouse)
-    default_schedules = False
     if len(schedules) == 0:
         wh = describe_warehouse(session, warehouse)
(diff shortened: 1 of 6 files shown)


app/crud/wh_sched.py
```diff
@@ -188,7 +188,18 @@ def create_table(cls, session, with_catalog_views=True):
         ).collect()
 
     @classmethod
-    def find_all(
+    def find_all(cls, session: Session, name: str) -> List["WarehouseSchedules"]:
+        """
+        Syntactic sugar to return all schedules for a given warehouse and weekday/weekend.
```
rymurr, line 193: i think the pydoc are swapped?

joshelser: whoopsies.

app/crud/wh_sched.py
```diff
@@ -277,6 +294,42 @@ def convert_time_str(time_str) -> datetime.time:
...
+    schedules = WarehouseSchedules.find_all(session, warehouse)
+    if len(schedules) == 0:
+        wh = describe_warehouse(session, warehouse)
+        wh.write(session)
+        schedules.append(wh)
+        wh2 = describe_warehouse(session, warehouse)
+        wh2.weekday = False
+        wh2.write(session)
+        schedules.append(wh2)
+    elif len(schedules) == 2 and all(not s.enabled for s in schedules):
```
joshelser, line 316: I'm going to have to come up with a better way to express this.

I didn't want to change the `internal.wh_schedules` table schema but I feel like that might be the correct thing to do when a user has edited the schedule. As this is, you can't create a schedule which sets the warehouse size to something other than the default (as it will be overwritten on update).

joshelser: fixed in latest patch.

joshelser:
Screenshot 2023-10-06 at 5 36 45 PM

screenshot of the "warning" (after modifying the schedule from the default)

rymurr (approved):

PR #454, opened 2023-10-09
Make schedule the first tab to mask streamlit-isms.

joshelser:


rymurr (approved):

vqmarkman (approved):
Thank you for doing it! I hope it will be more usable with this change.

PR #473, opened 2023-11-01
Update probes.py to avoid NULL in conditions

rymurr:


joshelser (approved):
I remember that `where false`, but have no recollection as to why/how that would've been dropped in the transition.

rymurr:
got merged as part of a different pr

PR #474, opened 2023-11-02
New init

rymurr:


app/crud/probes.py
```diff
@@ -124,12 +124,12 @@ def validate_notifications(cls, values: dict) -> dict:
     def validate_probe_condition(cls, values: dict) -> dict:
         session = get_current_session()
 
-        name = values.get("name")
+        values.get("name")
```
joshelser, line 127: Delete?

bootstrap/009_register.sql
```diff
@@ -0,0 +1,29 @@
+
+create table if not exists internal.external_id_table as (select randstr(16, random()) as external_id);
```
joshelser, line 2: Why make a new table vs. adding a row to `internal.config`?

bootstrap/100_final_perms.sql
```diff
@@ -13,3 +13,9 @@ grant select on all views in schema CATALOG to APPLICATION ROLE READ_ONLY;
 grant usage on all functions in schema TOOLS to APPLICATION ROLE READ_ONLY;
 grant usage on all functions in schema TOOLS to APPLICATION ROLE ADMIN;
 grant usage on all functions in schema TOOLS to APPLICATION ROLE PUBLIC_;
+
+-- perms for service account
+-- TODO should we limit to individual procs,views?
```
joshelser, line 18: Aside from the setup-related procs, I can't think of a reason that we'd want to limit access for the service role.

bootstrap/039_reference_management.sql
```diff
@@ -15,18 +15,12 @@ begin
...
        select system$set_reference(:ref_name, :ref_or_alias);
-        if (ref_name = 'OPSCENTER_API_INTEGRATION') then
-            insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'Running external funcitons setup proc.', :ref_or_alias);
-            call admin.setup_external_functions('opscenter_api_integration');
-        end if;
-        if (ref_name = 'OPSCENTER_SSO_API_INTEGRATION') then
-            insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'Running external funcitons setup proc.', :ref_or_alias);
-            call admin.setup_register_tenant_func();
-        end if;
+        insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'Running external funcitons setup proc.', :ref_or_alias);
```
joshelser, line 18:
```suggestion
        insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'Running external functions setup proc.', :ref_or_alias);
```

joshelser (approved):
Mostly nit-picky stuff. Only real question is about the external_id table vs. internal.config and even that's not going to make/break anything.

rymurr:
Unsurprisingly the cypress tests failed. I think its best to just disable for now.

joshelser:
> Unsurprisingly the cypress tests failed. I think its best to just disable for now.

agree. no sense in trying to update those while in flux.

bootstrap/100_final_perms.sql
rymurr: Only the principle of least privilege. Perhaps we only need to be able to modify the crud and settings, not make changes to other admin procs. Additionally, perhaps they don't want to give us access to all data, just the report card related stuff.

Either way, I want to defer to a later PR, when we know more

bootstrap/009_register.sql
rymurr: To keep it somewhat secret (eg not in the catalog view) and to ensrue it can never be overwritten w/o extra logic

app/crud/probes.py
rymurr: done

joshelser (dismissed):

bootstrap/009_register.sql
joshelser: I _alwasy_ forget about the config catalog view. Makes sense.

joshelser (approved):

PR #476, opened 2023-11-03
Convert minutes back to seconds when building ALTER WAREHOUSE

joshelser:


rymurr (dismissed):

joshelser:
just a test fix since last approval, merging.

PR #480, opened 2023-11-07
Missed the sortby on this one invocation.

joshelser:
When we get unlucky and the rows are misordered, it can cause the delete cleanup to fail.

```
Verify failed: Start time must be before finish time\\nid_val=\'d730ad7d6fa6400fb7168f28364dce6d\' name=\'integration_tests\' start_at=datetime.time(17, 0) finish_at=datetime.time(8, 0) size=\'Medium\' suspend_minutes=1 resume=True scale_min=1 scale_max=1 warehouse_mode=\'Standard\' comment=None weekday=True day=None enabled=True last_modified=NaT
```


joshelser:
The longer-winded explanation is that the `ignore_errors=True` kwarg on `verify_and_clean` will happily proceed forward when `schedules[0]` doesn't start at 00:00 and when `schedules[-1]` doesnt' finish at 23:59.

This will, however, fail at the end of the function when we call `i.validate(..)`.

I'm not sure if there's a better way to override the `batch_read()` classmethod on the BaseOpsCenterModel class with the WarehouseSchedules class? I could see someone else messing up this again in the future.

PR #482, opened 2023-11-09
Move the warehouse schedules table creation out of python

joshelser:
We moved the table creation into finalize_setup() because we cannot call a python procedure during the native app setup script. However, this created a different problem that this procedure needs to be called (e.g. the UI opened) before the user is able to see/use the catalog views.

Lift the table/view creation out of python and put it into SQL again.

### DIFF
--- a/bootstrap/011_warehouse_schedules.sql
+++ b/bootstrap/011_warehouse_schedules.sql
@@ -1,24 +1,49 @@
 
--- NB. Table is created in 090_post_setup.sql because you cannot call python procs from the setup.sql
-CREATE OR REPLACE PROCEDURE INTERNAL.CREATE_WAREHOUSE_SCHEDULES_VIEWS()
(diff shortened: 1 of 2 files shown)


joshelser:
Unrelated, I just noticed that I'm not able to run deploy.py and create a differently-named native app.

The Snowflake permissions SDK says that the manifest doesn't match the app's request:

Screenshot 2023-11-09 at 5 10 10 PM

```
The app JOSH_OPSCENTER_APP requested MANAGE WAREHOUSES, EXECUTE TASK, IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE, and EXECUTE MANAGED TASK privileges that are not defined in the app's manifest. Please fix the error and update the new app listing.
```

I don't know why it's only happening for me (and not also for CI).

rymurr (approved):

PR #484, opened 2023-11-13
Ensure that query monitors with duplicate conditions eventually fire

joshelser:
OpsCenter allows multiple query monitors to exist that have the same condition. A bug was identified by Jacques in which, after the first query monitor "fires", the subsequent query monitors duplicating the condition of the first QM, never fire. The expectation is that on the next "run" of the query monitor task execution, subsequent query monitors will "fire".

I solved this by lifting the logic to check the internal.probe_actions table (which records if a QM has already fired for a query) into a scalar UDF and embedded that in the case-statement which matches a query to a Query Monitor. This seemed like the "easiest" solution (but not necessarily the most elegant solution) -- this felt like the correct trade-off to me (simplicity over complexity). WRT performance, my expectation is that Snowflake would pull the partition of probe_actions into memory once (containing recent rows) and subsequent calls over the UDF would be "fast" (but I have not spend time to verify that is happening).

### DIFF
--- a/bootstrap/007_probes.sql
+++ b/bootstrap/007_probes.sql
@@ -125,6 +125,11 @@ BEGIN
     -- TODO: find a way to improve this so we're not double querying probes.
     -- We do it here so we don't have to transport objects from queries to sql text but it means
     -- there is a slight chance of seeing different actions for the probe between the first step and the second.
+    --
+    -- The actions CTE is doing two-levels of filtering to ensure each probe fires exactly one per matched query.
+    -- The addition clauses on top of the CONDITION in the case statements ensure probes with duplicate conditions
+    -- will each match. The final NOT IN clause of the where condition for the CTE will ensure that we don't re-trigger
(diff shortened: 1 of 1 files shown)


rymurr (dismissed):

jacques-n:
I think it would be better to do an outer join with a null check on the join condition columns. I wouldn't rely on the optimizer to do a single join as opposed to a row by row lookup. Maybe it would. 

It would also avoid creating a SQL function that is only used in one place.

joshelser:
> I think it would be better to do an outer join with a null check on the join condition columns. I wouldn't rely on the optimizer to do a single join as opposed to a row by row lookup. Maybe it would.

Thanks for your feedback! Agree with your points.

joshelser:
Thanks for Jacques for helping me with the specifics of this change.

I've managed to remove the scalar UDF and fix this bug, without doing a full rewrite of how we process query monitors.

rymurr (approved):

PR #485, opened 2023-11-14
Add a procedure to reset schedules for a warehouse back to the default

joshelser:
Also triggers the logic to update the internal "alter task" statements which was impossible to do even with admin.run_as_app.

Updates snowflake-based tests for this change (and removes python-based create table test code which should have been removed in an earlier commit)


joshelser:
Going to move ahead with this to get a release out today.

PR #498, opened 2023-11-27
Add new procedure to link OpsCenter to Sundeck

joshelser:
`admin.link_sundeck(text)` will accept an auth token for a Sundeck account and do the following:
* Store that token for future use
* Create all external functions against Sundeck

This procedure expects that the user has "opened" OpsCenter at least once and granted the requested permissions around API integrations.

Related fixes:
* The reference management callback procedure wrote two rows to the table instead of just one (as was likely intended)
* Removing a permission currently results in a sql error, so added some debugging to try to figure that out.
* Mocked python permissions API returned the wrong type, corrected to match the real python permissions API.

Posting as draft as the iteration cycle is slow. Still making sure this actually works as expected.

### DIFF
--- a/app/ui/permissions.py
+++ b/app/ui/permissions.py
@@ -53,5 +53,7 @@ def setup_permissions():
             None,
         )
     if not config.up_to_date():
+        st.info("Post-upgrade tasks are running in the background...")
         with connection.Connection.get() as conn:
             conn.call(f"{db}.ADMIN.FINALIZE_SETUP")
+            st.info("Upgrade successful!")
--- a/app/ui/permissions.py
+++ b/app/ui/permissions.py
@@ -53,7 +53,8 @@ def setup_permissions():
             None,
         )
     if not config.up_to_date():
-        st.info("Post-upgrade tasks are running in the background...")
-        with connection.Connection.get() as conn:
+        with connection.Connection.get() as conn, st.status(
+            "Post-upgrade tasks are running in the background..."
+        ) as status:
             conn.call(f"{db}.ADMIN.FINALIZE_SETUP")
-            st.info("Upgrade successful!")
+            status.update(label="Upgrade successful!")
--- a/app/ui/permissions.py
+++ b/app/ui/permissions.py
@@ -53,8 +53,5 @@ def setup_permissions():
             None,
         )
     if not config.up_to_date():
-        with connection.Connection.get() as conn, st.status(
-            "Post-upgrade tasks are running in the background..."
-        ) as status:
+        with connection.Connection.get() as conn:
             conn.call(f"{db}.ADMIN.FINALIZE_SETUP")
-            status.update(label="Upgrade successful!")
--- a/bootstrap/021_external_functions.sql
+++ b/bootstrap/021_external_functions.sql
@@ -344,6 +344,7 @@ create or replace procedure admin.connect_sundeck(token text)
     execute as owner
 as
 DECLARE
+    -- We could also use the internal.reference_management table for OpsCenter V2, but not clear if there is value in that.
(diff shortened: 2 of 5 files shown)


app/ui/permissions.py
```diff
@@ -53,5 +53,8 @@ def setup_permissions():
             None,
         )
     if not config.up_to_date():
-        with connection.Connection.get() as conn:
+        with connection.Connection.get() as conn, st.status(
+            "Post-upgrade tasks are running in the background..."
+        ) as status:
             conn.call(f"{db}.ADMIN.FINALIZE_SETUP")
+            status.update(label="Upgrade successful!")
```
rymurr, line 60: What has to still be run in finalize_setup? It would be good if a user didn't have to open up the UI every time the versions were updated. (can be part of a later PR)

bootstrap/021_external_functions.sql
```diff
@@ -337,3 +337,38 @@ begin
...
+    execute immediate 'create or replace function internal.get_ef_token() returns string as \'\\\'' || token || '\\\'\';';
+
+    -- Create all external functions that use the API Gateway and the auth token
+    call admin.setup_external_functions('opscenter_api_integration');
+
+    -- Success!
+    return (select object_construct());
+EXCEPTION
+    WHEN OTHER THEN
+        SYSTEM$LOG_ERROR(OBJECT_CONSTRUCT('error', 'Failed to connect to Sundeck. Please contact Sundeck for support.', 'SQLERRM', SQLERRM, 'SQLSTATE', SQLSTATE));
```
rymurr, line 372: is this as tight as we can make the error messages? We cant tell which step we broke at?

rymurr (commented):
LGTM but its currently hard to tell what is going to get dropped and what is going to stay and if we can make the experience tighter wrt only needing to open the UI once, makign sure we can communicate linking errors correctly etc

app/ui/permissions.py
joshelser: Aside: I should probably remove this and defer to a later change-- it's just "fixing" the blank page when finalize_setup is running post-upgrade.

From what I remember, the biggest piece of finalize_setup() are the table migration checks which take ~1-2minutes.

bootstrap/021_external_functions.sql
joshelser: Thanks for the suggestion. I can definitely make the error checking more granular.

app/ui/permissions.py
rymurr: that runs in the tasks too doesnt it? I don't remember, I just don't want to ahve a user open the UI after the first time

joshelser: I don't remember either, but I agree with your goal. I will take a look.

I think we will have to be extra careful of this going forward (ensuring any state in an unversioned schema is migrated automatically w/o the need to open the UI).

joshelser:
> its currently hard to tell what is going to get dropped and what is going to stay

It's not entirely clear to me, either, how many things we want to drop from the SQL procedures, especially if we may want to enable Snowflake SSO again later.

With the current direction we're going, I believe the following could be removed
* `admin.setup_register_tenant_func` procedure and related external function
* `admin.register_sundeck_account(username varchar, password varchar, role varchar, warehouse varchar)` procedure and related external function
* `internal.ef_register_service_account` external function
* Most of the contents of `crud/account.py` -- only using the `get_api_gateway_url` helper func right now (but security integration setup is included in there)

joshelser:
@rymurr I genuinely welcome your review here as this is "net-new" code instead of fixing bugs. I'm going to merge this so I can test out an OpsCenter v2, internally released.

PR #501, opened 2023-12-04
Better validation error messages for warehouse schedules

joshelser:
Found a few places where we did not include an error message on the assert statement. These are traveling to the user, so they should be terse yet specific.


joshelser:
Thanks to Vicky for noticing the original bad error message on warehouse size.

vqmarkman (approved):
Thanks for fixing other negative cases as well! Ship it.

rymurr (approved):

PR #502, opened 2023-12-04
Fix common.py error messages for query monitors

joshelser:
Vicky noticed that error messages generated on query monitor procedures still resulted in the word "probe" showing up.

Overload the Probe class with the QUERY_MONITOR entity type, and update procedures to use that instead of "PROBE".

### DIFF
--- a/app/crud/common.py
+++ b/app/crud/common.py
@@ -9,6 +9,7 @@ from .wh_sched import WarehouseSchedules, WarehouseAlterStatements
 _TYPES = {
     "LABEL": Label,
     "PROBE": Probe,
+    "QUERY_MONITOR": Probe,
     "WAREHOUSE_SCHEDULES": WarehouseSchedules,
     "WAREHOUSE_ALTER_STATEMENTS": WarehouseAlterStatements,
 }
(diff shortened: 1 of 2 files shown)


app/crud/common.py
```diff
@@ -9,6 +9,7 @@
 _TYPES = {
     "LABEL": Label,
     "PROBE": Probe,
+    "QUERY_MONITOR": Probe,
```
vqmarkman, line 12: "QUERY_MONITOR": Query monitor ?

rymurr (approved):

app/crud/common.py
joshelser: I don't want to change the entity type from being this format `ALL_CAPS` because this is critical to align between sql proc definition and the python code.

We do normalize the key by calling `lower()` on it before wrapping the error message, e.g.

```
return summarize_error(f"Failed to create {entity_type.lower()}", e)
```

We could remove the underscore if you think that's important.

vqmarkman: no important, ship it.

vqmarkman (approved):

PR #512, opened 2023-12-23
Correct issues in warehouse utilization calculation

jacques-n:
This fixes several problems:
- incorrect association of some queries as serverless when they are not (switch to using warehouse prefix matching)
- remove inconsistent exclusion of serverless queries from the unloaded cost calculation of warehouse utilization but being included in the loaded cost calculation
- ensuring that queries are delayed as long as metering history for match up
- avoiding ambiguous grouping in QUERY_WH_UTIL CTE that causes duplicate rows to be generated, causing overstatement.
- too strict of serverless naming as Snowflake sometimes uses just "COMPUTE_SERVICE_WH" without any additional suffix.
- disable the inclusion of serverless tasks in warehouse utilization reports. They seem to be frequently be 100% causing noise in the data (we can always revert in the future if some people are struggling with low serverless utilization).

More detail on why serverless mis-identification can cause problems...

It appears that in some accounts the WAREHOUSE_ID for serverless tasks starts much lower in some accounts than others. This means the IS_SERVERLESS_TASK(number) function is incorrectly reporting some serverless task queries as normal queries. In the warehouse utilization view, QUERY_WH_UTIL CTE then uses this decision which causes a mismatch on a warehouse_id basis between the two sides of the join (unloaded credits and loaded credits). Because unloaded cost is often over-estimated when looking at nested queries (fyi serverless tasks often use nested queries), we have an automatic correction system to ensure we never overcount utilization. However, because the join is mismatching, this correction isn't being applied in all cases. This can cause utilization to be overstated.

A note during debugging that made things harder... because query history and warehouse history reports are not using the same delayed visibility, today's utilization will almost always be overreported (since utilization uses the all but the last 45 minutes of query history but only includes metering history that is older than 3 hours). We should fix so the delays are consistent across views. For now, this patch addresses this problem by adding a 3 hour delay to the query history view to be inline with metering history view.


joshelser (approved):
We should remove the `internal.is_serverless_warehouse(number)` function, too, if it's not working as expected.

jacques-n:
> We should remove the `internal.is_serverless_warehouse(number)` function, too, if it's not working as expected.

Agreed but it is used elsewhere and didn't want to tackle that change in this ticket.

joshelser:
i will merge after https://github.com/sundeck-io/OpsCenter/actions/runs/7305538315 complete -- don't need to wait for another set of tests to run for just a whitespace change.

PR #516, opened 2024-01-10
 Add query_signature helper methods

anshuldata:


bootstrap/021_external_functions.sql
```diff
@@ -159,6 +159,111 @@ $$
...
+$$;
+
+create or replace function tools.signatures_match(db_first varchar, schema_first varchar, query_text_first varchar,
+            db_second varchar, schema_second varchar, query_text_second varchar)
+    returns boolean
+as
+$$
+    internal.wrapper_signatures_match(object_construct('database_first', db_first, 'schema_first', schema_first, 'query_text_first', query_text_first,
+    'database_second', db_second, 'schema_second', schema_second, 'query_text_second', query_text_second))
+$$;
```
joshelser, line 230: I suggest you put the `tools` UDFs into a different file (larger than 021). Could put them into 042_utils.sql or create a new file specific to signatures.

joshelser (changes requested):
Looks generally ok, just a request to move stuff around.

I think you also need to run `git submodule update --remote` to pick up the latest qtag update that was done upstream. I believe you unintentionally reverted the last qtag update.

bootstrap/021_external_functions.sql
anshuldata: done

anshuldata:
> I think you also need to run git submodule update --remote to pick up the latest qtag update that was done upstream. I believe you unintentionally reverted the last qtag update.

Thanks, fixed this

joshelser (approved):

rymurr (approved):

PR #517, opened 2024-01-18
Add a normalize sqlglot function

rymurr:
also updates sqlglot and adds more periods to the period_range functions


bootstrap/031_sqlglot.sql
```diff
@@ -9,24 +13,71 @@ as
...
 import sqlglot.expressions as exp
-def test(sql, database, schema):
+def test(sql, database, schema, exclude_cte):
+    if schema is None:
+        schema = ''
+    if database is None:
+        database = ''
     try:
-        return list(f"{database if table.catalog == '' else table.catalog}.{schema if table.db == '' else table.db}.{table.name}".upper() for
+        tables = list(f"{database if table.catalog == '' else table.catalog}.{schema if table.db == '' else table.db}.{table.name}".upper() for
```
joshelser, line 22: Do you have these backwards? `database` goes to `table.db` and `schema` with `table.catalog`?

glancing at https://sqlglot.com/sqlglot/expressions.html#Table feels like you might have these backwards.

bootstrap/031_sqlglot.sql
```diff
@@ -9,24 +13,71 @@ as
...
-$$;
+from sqlglot import parse_one
+import sqlglot.expressions as exp
+def transform(node, database, schema):
+    if schema is None:
+        schema = ''
+    if database is None:
+        database = ''
+    if isinstance(node, exp.Table) and node.name != '':
+        tbl = [node.catalog or database, node.db or schema, node.name]
```
joshelser, line 56: here too, should this be `node.db` ?

joshelser (approved):
Approving because I very-well might be misreading the sqlglot apidocs.

Only other thought is that if we push this down into app/python, we could write normal-looking pytests (without the sql procedure wrappers). Can be done later if these are used in earnest.

PR #521, opened 2024-01-19
forgot grants on new view

rymurr:


joshelser:
I'm surprised! I thought we re-granted all of the views in reporting after we recreated them. I guess I'm thinking of something else 🤔

PR #526, opened 2024-01-25
Revert "Banner to alert of unsupported Azure features"

rymurr:
Reverts sundeck-io/OpsCenter#524

Backend fixed and the Azure app now works fine


joshelser (approved):
Holler if you want another set of eyes on the azure app

PR #527, opened 2024-01-25
Some minor fixes while testing

rymurr:


joshelser (approved):
https://docs.snowflake.com/en/developer-guide/logging-tracing/tracing-python#adding-support-for-the-telemetry-package should we also list this as a package for the python procedures?

PR #531, opened 2024-01-29
Fully link Native app to Sundeck w/o opening UI

rymurr:


bootstrap/092_service_account_setup.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE OR REPLACE PROCEDURE admin.upgrade_check()
+returns varchar
+language sql
+as
+begin
+    let version varchar;
+    call internal.get_config('post_setup') into :version;
+    let setup_version varchar := (select internal.setup_version());
+    if (:version <> :setup_version) then
```
joshelser, line 25: It's possible that `:version` is `NULL` (user never opened native app). Should this be something like `if (:version is NULL or :version <> :setup_version)`?

joshelser (changes requested):

bootstrap/092_service_account_setup.sql
```diff
@@ -0,0 +1,29 @@
...
+as
+begin
+    execute immediate 'create or replace function internal.get_ef_url() returns string as \'\\\'' || url || '\\\'\';';
+    execute immediate 'create or replace function internal.get_tenant_url() returns string as \'\\\'' || web_url || '\\\'\';';
+
+    call internal.set_config('tenant_url', web_url);
+    call internal.set_config('url', url);
+    call admin.update_reference('OPSCENTER_API_INTEGRATION', 'ADD', SYSTEM$REFERENCE('API Integration', 'OPSCENTER_API_INTEGRATION', 'persistent', 'usage'));
+    call admin.connect_sundeck(token);
+    call admin.finalize_setup();
```
joshelser, line 14: Can we call this via the task so it happens in the backend to avoid this procedure taking a long time?

bootstrap/092_service_account_setup.sql
rymurr: yup, fixed

bootstrap/092_service_account_setup.sql
rymurr: as discussed Ive removed finalize and connect sundeck. The assumption being that these will be called by the linking service

joshelser (approved):

PR #533, opened 2024-01-29
Add telemetry library

rymurr:


app/ui/config.py
```diff
@@ -21,6 +21,10 @@ def up_to_date():
     return str(Config.get("post_setup")) == CurrentVersion.get()
 
 
+def get_current_and_expected_version():
+    return CurrentVersion.get(), Config.get("post_setup")
```
joshelser, line 25: Unused?

joshelser (approved):

app/ui/config.py
rymurr: currently yes. I added it b/c it was in v1 and I assume it will get used

joshelser: Fine to leave it in. Just didn't recall seeing it in the v1 PR and wanted to dbl check it was intentional.

PR #534, opened 2024-01-29
Ensure settings are created by default

rymurr:


bootstrap/090_post_setup.sql
```diff
@@ -366,5 +366,9 @@ call internal.has_config('tenant_url') into :has_tenant_url;
 if (:has_url or :has_tenant_url) then
     call internal.start_user_limits_task();
 end if;
+call internal.set_config('compute_credit_cost', '2.0');
+call internal.set_config('serverless_credit_cost', '3.0');
+call internal.set_config('storage_cost', '40.0');
+call internal.set_config('default_timezone', 'America/Los_Angeles');
```
joshelser, line 372: I think the original goal was that we pull the timezone from properties, pulling `value` from

```
>SHOW PARAMETERS LIKE 'TIMEZONE';
+----------+---------------------+---------------------+-------+-------------+--------+
| key      | value               | default             | level | description | type   |
|----------+---------------------+---------------------+-------+-------------+--------|
| TIMEZONE | America/Los_Angeles | America/Los_Angeles |       | time zone   | STRING |
+----------+---------------------+---------------------+-------+-------------+--------+
```

joshelser (changes requested):

bootstrap/090_post_setup.sql
rymurr: well, we cant access `show parameters` from native app. So eithe rthe service account sets it as part of the hook up process or we set it to some default. Thoughts?

rymurr: I think its probably best to set it in the service account linking stage, WDYT?

joshelser (approved):

bootstrap/090_post_setup.sql
joshelser: (I don't really feel strongly if it's empty vs. a real timezone)

rymurr: good shout. Ive set it to a default and will leave setting it to you in the linking script

joshelser (approved):

PR #536, opened 2024-01-30
Minor bugfix from linking native app

joshelser:
1. Create the UPGRADE_CHECK task outside of FINALIZE_SETUP
2. s/internal.setup_version/internal.get_version/

### DIFF
--- a/bootstrap/090_post_setup.sql
+++ b/bootstrap/090_post_setup.sql
@@ -61,13 +61,6 @@ CREATE OR REPLACE TASK TASKS.SFUSER_MAINTENANCE
     AS
     CALL INTERNAL.refresh_users();
 
-CREATE OR REPLACE TASK TASKS.UPGRADE_CHECK
-    SCHEDULE = '1440 minute'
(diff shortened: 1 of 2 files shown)


joshelser:
Ah, drat. I had convinced myself we could make serverless tasks outside of finalize setup. That's why you put it in there. Can come up with something else.

joshelser:
a third one:
``` 
000904 (42000): Uncaught exception of type 'STATEMENT_ERROR' on line 13 at position 4 : SQL compilation error: error line 1 at position 39
invalid identifier 'WEB_URL'
```

maybe on the `internal.set_config(..)` calls

joshelser:
Moved it to draft since I don't have the task creation working locally.

rymurr:
I think the finalize task will have to be created separately by the service account

joshelser:
> I think the finalize task will have to be created separately by the service account

Thanks. Yeah, that's what I started doing and then ran into the last bug. Will get that sorted and move this out of draft.

PR #540, opened 2024-01-30
More bugfixes on new linking process

joshelser:
* s/setup_version/get_version/
* Moves tasks.upgrade_check into admin.finalize_setup_with_service_account()
* Removes dead code from admin.update_reference()
* Adds an overloaded admin.finalize_setup_with_service_account() to specify an alternate API integration
* Avoids use of admin.update_reference() in finalize_setup_with_service_account() since it expects the internal.get_ef_token() function to already be created.


bootstrap/092_service_account_setup.sql
```diff
@@ -1,15 +1,59 @@
...
+    select system$set_reference(:ref_name, :ref_val);
+    insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'finalize_setup_from_service_account', :ref_val);
+end;
+
 CREATE OR REPLACE PROCEDURE admin.finalize_setup_from_service_account(url varchar, web_url varchar)
 RETURNS varchar
 LANGUAGE sql
 as
+declare
+    integration_name text default (select 'OPSCENTER_SUNDECK_EXTERNAL_FUNCTIONS');
```
rymurr, line 35: I don't think this is the right name. That doesn't match the manifest or the api integration we suggest creating in the help script

joshelser:
Will cherry-pick this to main after I get it happy on v1

PR #551, opened 2024-02-08
Store autoscaling available (or not) as config. Make wrapper that ret…

joshelser:
…urns warehouses and size


rymurr (approved):

joshelser:
I don't know why cypress is yelling all of a sudden. It seems to be flakiness in the tests themselves (a much smaller set of failures on re-run). Looking at the UI I deployed locally do not indicate any problems.

PR #556, opened 2024-02-19
Improve signup performance

rymurr:


bootstrap/021_external_functions.sql
```diff
@@ -357,13 +362,6 @@ BEGIN
             headers = (\'sndk-token\' = \'' || token || '\')
             as \'' || url || '/extfunc/update_usage_stats\';
 
-            create or replace external function internal.ef_register_service_account(service_account object)
```
rymurr, line 360: unused ef

bootstrap/021_external_functions.sql
```diff
@@ -413,20 +401,7 @@ END;
 
 CREATE OR REPLACE PROCEDURE admin.setup_sundeck_tenant_url(url string, token string) RETURNS STRING LANGUAGE SQL AS
 BEGIN
-    execute immediate 'create or replace function internal.get_tenant_url() returns string as \'\\\'' || url || '\\\'\';';
```
rymurr, line 416: this is done above

bootstrap/021_external_functions.sql
```diff
@@ -413,20 +401,7 @@ END;
...
     execute immediate 'create or replace function internal.get_ef_token() returns string as \'\\\'' || token || '\\\'\';';
-
-    MERGE INTO internal.config AS target
```
rymurr, line 419: this is done above

bootstrap/039_reference_management.sql
```diff
@@ -16,7 +16,6 @@ begin
     when 'ADD' then
        select system$set_reference(:ref_name, :ref_or_alias);
        insert into internal.reference_management (ref_name, operation, ref_or_alias) values (:ref_name, 'Running external functions setup proc.', :ref_or_alias);
-       call admin.setup_register_tenant_func();
```
rymurr, line 19: unused ef

bootstrap/090_post_setup.sql
```diff
@@ -348,6 +348,7 @@ call internal.maybe_set_config('default_timezone', 'America/Los_Angeles');
 let has_autoscaling boolean;
 call internal.account_has_autoscaling() into :has_autoscaling;
 call internal.set_config('autoscaling_available', :has_autoscaling);
+CALL admin.setup_external_functions('opscenter_api_integration');
```
rymurr, line 351: do this async

bootstrap/092_service_account_setup.sql
```diff
@@ -16,13 +16,14 @@ begin
         AS
         CALL ADMIN.UPGRADE_CHECK();
     grant MONITOR, OPERATE on TASK TASKS.UPGRADE_CHECK to APPLICATION ROLE ADMIN;
+    alter task TASKS.UPGRADE_CHECK resume;
```
rymurr, line 19: we didn't start the task!

bootstrap/092_service_account_setup.sql
```diff
@@ -16,13 +16,14 @@ begin
...
 
     call internal.set_config('tenant_url', :web_url);
     call internal.set_config('url', :url);
 
     -- Bind the given reference ID to the 'OPSCENTER_API_INTEGRATION' reference. Must match the reference in manifest.yml
     -- differs from v1 in that all external functions are not re-created by update_reference()
-    call admin.update_reference('OPSCENTER_API_INTEGRATION', 'finalize_setup_from_service_account', :api_integration_ref_id);
+    call admin.update_reference('OPSCENTER_API_INTEGRATION', 'ADD', :api_integration_ref_id);
```
rymurr, line 26: I thikn the update ref command was always broken?

bootstrap/021_external_functions.sql
```diff
@@ -325,6 +325,11 @@ END;
 
 CREATE OR REPLACE PROCEDURE admin.setup_external_functions(api_integration_name string) RETURNS STRING LANGUAGE SQL AS
 BEGIN
+    let hasUrl varchar;
```
rymurr, line 328: skip if ran too early...will get ran again soon

app/manifest.yml
```diff
@@ -6,7 +6,7 @@ artifacts:
   default_streamlit: admin.opscenter
 
 configuration:
-  trace_level: ON_EVENT
+  trace_level: ALWAYS
```
rymurr, line 9: temp while testing

bootstrap/092_service_account_setup.sql
joshelser: system$set_reference has to be called by a user and not the native app, IIRC. We should replace this with an insert to `internal.reference_management` or add some new "operation".

bootstrap/092_service_account_setup.sql
joshelser: See the comment above?

> Does not start the task -- the first time the task runs, finalize_setup() will start the task

bootstrap/092_service_account_setup.sql
rymurr: ahhhh...ok. I will just add the insert here

joshelser: Yeah, something that does an `insert` seems reasonable. Taking a step back, I'm not really sure if we had a goal with the reference_management table (or if we just copied the public examples), but I don't think we have anything that actually reads from it today.

bootstrap/092_service_account_setup.sql
```diff
@@ -25,7 +25,8 @@ begin
 
     -- Bind the given reference ID to the 'OPSCENTER_API_INTEGRATION' reference. Must match the reference in manifest.yml
     -- differs from v1 in that all external functions are not re-created by update_reference()
-    call admin.update_reference('OPSCENTER_API_INTEGRATION', 'finalize_setup_from_service_account', :api_integration_ref_id);
+   insert into internal.reference_management (ref_name, operation, ref_or_alias) values ('OPSCENTER_API_INTEGRATION', 'Running external functions setup proc.', :api_integration_ref_id);
+    call admin.connect_sundeck(:token);
```
joshelser, line 29: `connect_sundeck` returns an object which contains some specific error information. It would be good to change this proc to `RETURNS OBJECT` and return the result from `admin.connect_sundeck`

rymurr: done

joshelser (changes requested):
One more comment (wrong config key).

Do we need to support the 3-arg and the 4-arg versions of `admin.finalize_setup_from_service_account` for a short time to avoid a complicated compatibility dance? i.e. leave 3-arg definition in place, release 4-arg definition, start calling 4-arg procedure, drop 3-arg definition.

bootstrap/021_external_functions.sql
```diff
@@ -325,6 +325,11 @@ END;
...
+    call internal.get_config('ef_url') into :hasUrl;
```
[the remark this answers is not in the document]
rymurr: good eye

rymurr:
> One more comment (wrong config key).
> 
> Do we need to support the 3-arg and the 4-arg versions of `admin.finalize_setup_from_service_account` for a short time to avoid a complicated compatibility dance? i.e. leave 3-arg definition in place, release 4-arg definition, start calling 4-arg procedure, drop 3-arg definition.

I didn't explicitly drop the 3-arg version. We will have to clean it up in another commit I guess. I don't expect to be able to get rid of it any time soon though...

joshelser (approved):

rymurr:
to be honest I forgot it was a versioned schmea. i'll re-=add it. thanks.

joshelser (approved):

PR #560, opened 2024-02-23
Share REPORTING.SUNDECK_QUERY_HISTORY view with user.

joshelser:
* Shared view must existing in a versioned schema. ACCOUNT_USAGE is an unversioned schema and contains non-user-facing views, so I put it into REPORTING schema.
* Prior to connecting the native app to Sundeck, the view is empty.
* To avoid creating a new versioned schema (or enumerating all other VIEWS in the REPORTING schema), the ADMIN and READ_ONLY application roles can always read this view (even if Sundeck is not-yet linked).
* If sundeck is not linked, the view is empty.
* Creates an internal.regions view based on the data in the streamlit app Python codebase to filter SNOWFLAKE_REGION and SNOWFLAKE_CLOUD
* For the marketplace-listing account, I believe I have the necessary changes for deploy.py to succeed once the SUNDECK database is updated but did not modify that account by hand to test
* Verified that I could not access data outside of my `CURRENT_ACCOUNT()`


bootstrap/010_helpers.sql
```diff
@@ -149,3 +149,44 @@ $$
 
   return str;
 $$;
+
+create or replace view internal.regions as
```
rymurr, line 153: do we not have access to `show regions`?

bootstrap/060_sharing.sql
```diff
@@ -0,0 +1,26 @@
...
+    -- Remove the REGION_GROUP when present from current_region()
+    select UPPER(region) from internal.regions where upper(SF_REGION) = upper(SPLIT_PART(CURRENT_REGION(), '.', -1))
+$$;
+
+-- Access to the SUNDECK database is given to application_package in deploy.py
+-- SHARING.GLOBAL_QUERY_HISTORY is already filtered on SNOWFLAKE_ACCOUNT_LOCATOR = CURRENT_ACCOUNT()
+CREATE OR REPLACE VIEW REPORTING.SUNDECK_QUERY_HISTORY
+   AS SELECT * from SHARING.GLOBAL_QUERY_HISTORY WHERE
+     UPPER(SUNDECK_ACCOUNT_ID) = (select UPPER(value) from internal.config where key = 'tenant_id') and
+     UPPER(SNOWFLAKE_REGION) = (select internal.parse_region()) and
```
rymurr, line 25: we can't use `internal` functions for the filtering of these views. They can be changed. We have to use `CURRENT_REGION` and `CURRENT_ACCOUNT` here

deploy/deploy.py
```diff
@@ -167,6 +174,36 @@ def _install_or_update_package(
...
+    print("Running grants to database SUNDECK for application package.")
+    # Grant REFERENCE_USAGE on SUNDECK to the application package
+    # Grant access to the view from sundeck
+    cur.execute(
+        f"""
+    BEGIN
+        CREATE SCHEMA IF NOT EXISTS "{APPLICATION_PACKAGE}".SHARING;
+
+        show databases;
+        let db_exists boolean := (SELECT "name" is not null FROM TABLE(RESULT_SCAN(LAST_QUERY_ID())) WHERE "name" = '{sundeck_db}');
```
rymurr, line 187: why not just do a `create if not exists`?

deploy/deploy.py
```diff
@@ -167,6 +174,36 @@ def _install_or_update_package(
...
+        if (db_exists) then
+            GRANT REFERENCE_USAGE ON DATABASE {sundeck_db} TO SHARE IN APPLICATION PACKAGE "{APPLICATION_PACKAGE}";
+            -- Create a view in the application package that is filtered to CURRENT_ACCOUNT(). It is critical that the
+            -- filtering is done here to ensure a user only sees their own history.
+            CREATE OR REPLACE VIEW "{APPLICATION_PACKAGE}".SHARING.GLOBAL_QUERY_HISTORY AS SELECT * FROM {sundeck_db}.INTERNAL.GLOBAL_QUERY_HISTORY
+                WHERE UPPER(SNOWFLAKE_ACCOUNT_LOCATOR) = UPPER(CURRENT_ACCOUNT());
```
rymurr, line 193: its not clear to me how this works...we are executing a statement with `CURRENT_ACCOUNT` in our account. This will only share our data?

joshelser: > we are executing a statement with CURRENT_ACCOUNT in our account. This will only share our data?

I had expected the current_account would be late-bound to the current_account when the view is queried, but I have not actually tested that is happening.

bootstrap/010_helpers.sql
joshelser: Not sure off the top of my head. I remembered we had a similar enumeration in the streamlit Python code and lifted it here.

rymurr: yeah, I think we were being lazy in teh streamlilt code. Wondering if we really need to be

deploy/deploy.py
rymurr: might be worth spinnign up an app and sharing w/ anothe rone of our accounts. I don't like that the security stuff is split across two views and I think it would be good to test, verify and consoilidate down

bootstrap/010_helpers.sql
joshelser: Pushed a view, built from `show regions`, into the application package which appears to do the trick.

deploy/deploy.py
joshelser: Pushed down region and cloud via the context functions into the application package view. The tenant_id filter is still happening in the setup script, but I'm not sure of a way around that. An adventurous user could potentially see queries executed in their Snowflake account by a different Sundeck account. I could imagine scenarios where that is "very bad". I'll take another look and see if I can come up with a way to push the Sundeck account info down, too.

joshelser:
Also, I created listing in our Sandbox account called "Josh Opscenter". I have it shared to a number of consumer accounts in AWS and Azure to verify things work as expected -- the only difference is that I have a static copy of data (CTAS) being shared rather through the App Package rather than reading from the "real view". Thanks for Vicky for walking through it with me, as well.

deploy/deploy.py
```diff
@@ -167,6 +174,59 @@ def _install_or_update_package(
...
+            SUNDECK_START_TIME timestamp_ntz,
+            SUNDECK_ACCOUNT_ID text,
+            ACTIONS_EXECUTED text,
+            SCHEMA_ONLY_REQUEST boolean);
+
+        GRANT REFERENCE_USAGE ON DATABASE {sundeck_db} TO SHARE IN APPLICATION PACKAGE "{APPLICATION_PACKAGE}";
+
+        SHOW REGIONS;
+        CREATE OR REPLACE TABLE "{APPLICATION_PACKAGE}".SHARING.REGIONS(snowflake_region text, cloud text, region text) AS
+            SELECT "snowflake_region", "cloud", "region" from table(result_scan(last_query_id()));
```
joshelser, line 212: It looks like we have a disconnect with Azure regions and the region field that Sundeck is exporting. `eastus2` vs. `east-us-2`.

joshelser: I've filed a ticket internally for us and corrected the region data in the view.

deploy/deploy.py
```diff
@@ -167,6 +174,62 @@ def _install_or_update_package(
         )
 
 
+def _grant_sundeck_db_access(cur, sundeck_db: str):
+    print("Running grants to database SUNDECK for application package.")
+    # Grant REFERENCE_USAGE on SUNDECK to the application package
+    # Grant access to the view from sundeck
+    cur.execute(
+        f"""
+    BEGIN
```
rymurr, line 183: can we move this to a sql file? Its hard to read enbedded in a python script

deploy/deploy.py
```diff
@@ -209,6 +272,7 @@ def execute(
     deployment: str = "prod",
     install: bool = True,
     skip_package: bool = False,
+    sundeck_db: str = "SUNDECK",
```
rymurr, line 275: i thikn if sundeck_Db is `None` the installation of hte share shoudl be skipped (smoothly)

joshelser: np. Is the creation of a dummy view as this PR is currently doing sufficient? I could also push down a check into the setup script which detects if the app package has the view we expect and avoid creating the REPORTING view.

joshelser: if `sundeck_db is None`, we will run a trimmed down view creation that has no data so the setup script can be just assume the `global_query_history` view is always present.

rymurr (approved):

PR #563, opened 2024-02-27
Update 020_warehouse_utilzation.sql

rymurr:


jacques-n:
Casting wrong side of comparison I think.

bootstrap/020_warehouse_utilzation.sql
```diff
@@ -79,12 +79,12 @@ as $$
...
-        where period between INPUT_START_DATE and INPUT_END_DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end
+        where period between INPUT_START_DATE::DATE and INPUT_END_DATE::DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end
         group by 1
 	)
 
     select cast(d.date as timestamp) as period, utilization from
     internal.dates d
     left outer join util u on d.date = u.period
-    where d.date between INPUT_START_DATE and INPUT_END_DATE
+    where d.date between INPUT_START_DATE::DATE and INPUT_END_DATE::DATE
```
joshelser, line 89: `d.date` here is a timestamp (looking at the definition of `internal.dates`). I'm not 100% certain, but am guessing that you are trying to avoid comparisons of `timestamp between date and date` and `date between timestamp and timestamp` here to make the range inclusive?

joshelser (approved):

PR #570, opened 2024-02-29
centralize python stored proc

rymurr:


rymurr:
This didn't make a meaningful difference in testing :-(

PR #577, opened 2024-03-05
Add external function for broker catalog

anshuldata:
* usage "select * from Table(list_brokers())"

* Tested with privately deployed app
* Column name same as DDL

Screenshot 2024-03-05 at 5 17 26 PM

### DIFF
--- a/bootstrap/044_brokers.sql
+++ b/bootstrap/044_brokers.sql
@@ -1,0 +1,17 @@
+
+
+create or replace function admin.list_brokers()
+    returns TABLE (flow_name VARCHAR, flow_is_default VARCHAR, flow_default_warehouse VARCHAR, flow_hostname VARCHAR, flow_last_modified_at VARCHAR, flow_ddl VARCHAR, flow_results_path VARCHAR, flow_comment VARCHAR)
+as
+$$
+    SELECT
+    value:name::string,
+    value:isDefault::string,
+    value:defaultWarehouse::string,
+    value:flowUrl::string,
+    value:lastModifiedDate::string,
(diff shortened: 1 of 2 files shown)


pravindra (approved):

pravindra (approved):

scgkiran (approved):

bootstrap/044_brokers.sql
```diff
@@ -0,0 +1,17 @@
+
+
+create or replace function admin.list_brokers()
+    returns TABLE (flow_name VARCHAR, flow_is_default VARCHAR, flow_default_warehouse VARCHAR, flow_hostname VARCHAR, flow_last_modified_at VARCHAR, flow_ddl VARCHAR, flow_results_path VARCHAR, flow_comment VARCHAR)
```
joshelser, line 4: The `flow_` prefix on all of these columns is a little odd (given we're in a `list_brokers` procedure). Low-pri: I think you could drop the prefix on all of them (`flow_name` -> `name`, `flow_is_default` -> `is_default`).

pravindra: I am good with dropping the prefix but pls change the column names in the DDL too [here](https://github.com/sundeck-io/sundeck/blob/main/services/query_service/sundeck/sundeck_flow_handler.go#L355) so that they are in-sync.

anshuldata:
> @anshuldata don't forget to cherry-pick this change to the `v1` branch, as well. Thanks!

I will do it after next prod deploy for SaaS changes to go in. I will do after that
Thanks for reminder

PR #580, opened 2024-03-08
Update 031_sqlglot.sql

rymurr:
improve parsing speed


joshelser (approved):
Neat. Just read about the vectorized python UDF support.

How much of the expected change is from the vectorization vs. the `optimizer.qualify` call instead of rebuilding the query from the AST?

PR #582, opened 2024-03-09
SQL stored procs to replace python

rymurr:


bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,55 @@
...
+    let set_cmds varchar := 'set dest.' || keys[0] || ' = src.' || keys[0];
+    for i in 1 to (array_size(:keys) - 1) do
+        cols := cols || ',' || keys[i];
+        vals := vals || ',' || 'obj[\'' || keys[i] || '\'] as ' || keys[i];
+        ins_cols := ins_cols || ', src.' || keys[i];
+        set_cmds := set_cmds || ', dest.' || keys[i] || ' = src.' || keys[i];
+    end for;
+    let stmt varchar := 'merge into internal.' || :tbl || ' dest using (select ' || :vals || ' from (select parse_json(?) as obj)) src on dest.' || :key || ' = src.' || :key|| ' when matched then update ' || :set_cmds || ' when not matched then insert (' || :cols || ') values (' || :ins_cols || ')' ;
+    execute immediate :stmt using (obj);
+    return :stmt;
```
joshelser, line 53: Does the caller get a reasonable error message if this fails or is it opaque (because of the `execute immediate` or the `execute as owner`)?

If the caller gets an icky error, I think it would be to add an `EXCEPT` and try to return something sensible... stringified version of the object we're trying to write or the merge statement we constructed.

In either case, it would be helpful to add a debug log message.

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,55 @@
...
+        for rr in cc do
+            if (not rr.result) then
+                return 'Error in validation: ' || rec.message;
+            end if;
+        end for;
+    exception
+        when others then
+            return 'Error in validation: ' || rec.message || ' exception ' || sqlerrm;
+    end;
+    end;
```
joshelser, line 28: Double `end;`?

bootstrap/004_labels.sql
```diff
@@ -1,9 +1,10 @@
 
--- we should remove this and the other creaste statement in favour of driving the creation of htese tables from the python model
 CREATE TABLE INTERNAL.LABELS if not exists (name string, group_name string null, group_rank number, label_created_at timestamp, condition string, enabled boolean, label_modified_at timestamp, is_dynamic boolean, label_id string);
 
 CREATE TABLE INTERNAL.PREDEFINED_LABELS if not exists (name string, group_name string null, group_rank number, label_created_at timestamp, condition string, enabled boolean, label_modified_at timestamp, is_dynamic boolean);
 
+create table if not exists internal.validate_labels(sql text, message text, simple boolean);
```
joshelser, line 6: Should be `create or replace table`? After each upgrade, we would re-insert the same rows.

joshelser (changes requested):

PR #584, opened 2024-03-13
exclude serverless in utilization report

rymurr:


vqmarkman (dismissed):

bootstrap/020_warehouse_utilzation.sql
```diff
@@ -79,7 +79,7 @@ as $$
     util as (
         select date_trunc('day', period) as period, sum(loaded_cc * (select any_value(value) from ccost)) as cost, iff(sum(loaded_cc) = 0,null, sum(unloaded_cc)/sum(loaded_cc)) as utilization
         from reporting.warehouse_daily_utilization
-        where period::DATE between INPUT_START_DATE and INPUT_END_DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end
+        where period::DATE between INPUT_START_DATE and INPUT_END_DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end and warehouse_id < 10000
```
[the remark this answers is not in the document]
rymurr: good idea, fixed

bootstrap/020_warehouse_utilzation.sql
```diff
@@ -79,7 +79,7 @@ as $$
     util as (
         select date_trunc('day', period) as period, sum(loaded_cc * (select any_value(value) from ccost)) as cost, iff(sum(loaded_cc) = 0,null, sum(unloaded_cc)/sum(loaded_cc)) as utilization
         from reporting.warehouse_daily_utilization
-        where period::DATE between INPUT_START_DATE and INPUT_END_DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end
+        where period::DATE between INPUT_START_DATE and INPUT_END_DATE and case when INPUT_WAREHOUSE_NAME is null then true else warehouse_name = INPUT_WAREHOUSE_NAME end and not internal.is_serverless_warehouse(warehouse_name)
```
joshelser, line 82: If `warehouse_name is null`, then the UDF is going to return `NULL` and, I think, mess up the condition. I think you would have to push it into the `CASE`

```
.. and 
case when INPUT_WAREHOUSE_NAME is null then 
   not internal.is_serverless_warehouse(warehouse_id)
else
  warehouse_name = INPUT_WAREHOUSE_NAME and not internal.is_serverless_warehouse(warehouse_name)
end
```

And, writing this out.. should the `true` branch of the `CASE`

rymurr: warehouse_utilization will never have a `NULL` name, are you maybe thinking abuot enriched queyr history?

joshelser: Yes, likely so 😅. I remember Jacques last PR having to work around this but I didn't dig up the context.

joshelser (approved):

PR #589, opened 2024-03-19
Fix cypress tests

rymurr:


rymurr:
The test I was aimin gto fix is fixed. Still have the one from `Home` failing. I'll merge once all CI completes

PR #591, opened 2024-03-20
SQL stored procs for labels to replace python procs

joshelser:
* Split up label tests into separate files


joshelser:
Non-scientific results show create label latency on the low end around 500ms and on the upper end around 2.5s. I would imagine we can chop some of that off by consolidating the number of "validation rows". I wasn't focused on that in this pass (more focused on correctness and maintainable tests).

joshelser:
Jacques mentioned that he wanted to take a look at this change, so I've tagged him for review.

joshelser:
FWIW, there wasn't much in terms of python tests we added that weren't already encapsulated in the label test cases (largely, because we didn't write pytests that required Snowflake -- just exercising the pydantic validations).

jacques-n:
Hey, can you separate this into two PRs? 

PR1: Moving runtime validation logic from python to SQL. 
PR2: Separating python tests into multiple files (with virtually no code changes).

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,63 @@
...
+end;
+$$;
+
+create or replace procedure admin.write(obj object, tbl text, key text)
+returns text
+language sql
+execute as owner
+as
+$$
+begin
```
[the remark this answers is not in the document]
rymurr: This was something I initially prototyped. I agree with your comment that the code is impenetrable but I dont follow your suggestions. It is indeed very hard to read but I don't know a better way to generate a generic merge statement.

In your opinion:
Is it a matter of re-organising the code into UDFs to be more clear?
do we not use a defined schema for the storage tables but intsead write a single variant column?

bootstrap/004_labels.sql
```diff
@@ -114,38 +118,116 @@ SELECT *,$$;
...
 
+create or replace procedure internal.set_label_conditions()
+returns text
+language sql
+execute as owner
+as
+$$
+begin
+    -- Ungrouped labels
+    insert into internal.validate_labels (sql, message, simple, create_only) values
```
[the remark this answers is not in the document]
rymurr: I am not sure I understand this feedback either (maybe my english parser is broken) Maybe a short call would make it easier?

test/unit/test_label_procs.py
```diff
@@ -208,9 +208,20 @@ def test_smoke_dynamic_grouped_label(conn, timestamp_string):
...
+    # Create a dynamic label
+    old_dyn_label = generate_unique_name("old_dyn_label", timestamp_string)
+    sql = f"call ADMIN.CREATE_LABEL(NULL, '{old_dyn_label}', NULL, 'QUERY_TYPE', TRUE);"
+    assert run_proc(conn, sql) is None, "ADMIN.CREATE_LABEL did not return NULL value!"
 
     label = generate_unique_name("label", timestamp_string)
-    sql = statement.format(label=label)
+    sql = statement.format(
+        label=label, old_label=old_label, old_dyn_label=old_dyn_label
+    )
```
joshelser, line 224: The python label implementation validated the conditions first and then verified existing labels against the input (e.g. cannot create a label with the same name as one that exists, a label has to exist with the name being updated).

I can separate this change out to its own PR, but hopefully this is clear enough.

joshelser:
Pushed a few commits:

* `admin.write(..)` now incrementally builds up the "pieces" from the incoming `object`, to then combine into the skeleton of the `MERGE` statement. This is much more readable to me.
* Tried to capture the obvious validation (is not, is not null) as UDFs that generate the `condition`. 
* I changed over the validations for ungrouped labels to creating a single object, and updated `admin.validate` to use those.

I didn't migrate the grouped labels and dynamic labels validations (so those tests will fail). I wanted to see what you two thought.

rymurr (commented):
I wonder if we can split this into a few PRs. Just having trouble reviewin geveyrhting,

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,162 @@ SELECT *,$$;
...
+AS
+$$
+    tools.templatejs('(select count(\*) = 0 from {table_name} where {null_column_name} is null and {uniq_column_name} = f:{uniq_column_name})',
+        {'table_name': table_name, 'uniq_column_name': uniq_column_name, 'null_column_name': null_column_name})
+$$;
+
+-- Ungrouped labels
+-- simple=true and create_only=false as the defaults
+CREATE OR REPLACE VIEW INTERNAL.VALIDATE_LABELS AS
+    select $1 as obj from (
```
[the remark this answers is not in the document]
joshelser: Frustratingly, not with the UDFs as the code currently has (to my knowledge). You can only use values when each value evaluates to a constant expression at compilation time.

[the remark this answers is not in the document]
joshelser: Ok, yes, Snowflake can collapse the UDF to a constant expression so I could switch this back to using `VALUES` but you can't use `object_construct()` in a VALUES statement.

joshelser:
@rymurr feedback from friday. Notably, I dropped grouped and dynamic label stuff from the current PR to reduce the size of it. One addition that we didn't discuss was that I also dropped the overloaded CREATE_LABEL and UPDATE_LABEL procedures (making `is_dynamic default false` lets us avoid the same blocks of code twice).

The UPDATE_LABEL procedure is still complicated -- I feel like the clean way to address complexity there is to make separate procedures for other labels, rather than overloading into singular procedures.

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,74 @@
+
+create or replace procedure admin.validate(obj object, validation_table text, is_create boolean)
+returns text
+language sql
+execute as owner
+as
+$$
+begin
+    -- Select the validation rows. In a create, choose all. In an update, omit rows which are for create_only.
+    let validation_query text := (select tools.templatejs(
```
rymurr, line 10: in the interest of fewer calls to js I am not sure if we need this. I would maybe do smoething like:
```
let validation_query text := 'select * from identifier('internal.validate_' || ?) where iff(?, true, NOT COALESCE(obj[\'create_only\'], FALSE))'
let rs resultset := (execute immediate :validation_query using (:validation_table, :is_create));
```

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,74 @@
...
+create or replace procedure admin.write(obj object, tbl text, key text)
+returns text
+language sql
+execute as owner
+as
+begin
+    let keys array := object_keys(:obj);
+
+    -- The USING portion of the merge statement.
+    let using_select_cols array := (select array_agg(tools.templatejs('obj[\'{col}\'] as {col}', {'col': value})) from table(flatten(input=>:keys)));
```
rymurr, line 50: these usages of templatejs seem superflouous. Can we remove them and save on calls to js?

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,122 @@ SELECT *,$$;
...
+$$;
+
+
+CREATE OR REPLACE PROCEDURE VALIDATION.VALID_LABEL_CONDITION(input varchar)
+    RETURNS BOOLEAN
+    LANGUAGE SQL
+AS
+BEGIN
+    let c varchar := (select parse_json(:input):condition);
+    let stmt varchar := (select 'select case when ' || :c || ' then true else false end from reporting.enriched_query_history limit 1');
```
rymurr, line 153: this coul duse `execute immediate stmt using (c)` which would simplify the concat above

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,122 @@ SELECT *,$$;
...
+CREATE OR REPLACE PROCEDURE VALIDATION.IS_VALID_LABEL_NAME(input varchar)
+    RETURNS BOOLEAN
+    LANGUAGE SQL
+    comment='Validates that the label name does not exist as a column in the reporting.enriched_query_history table'
+AS
+BEGIN
+    let n varchar := (select parse_json(:input):name);
+    let stmt varchar := (select 'select ' || :n || ' from reporting.enriched_query_history where false');
+    execute immediate stmt;
+    return false;
```
rymurr, line 167: same

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,122 @@ SELECT *,$$;
...
+
+CREATE OR REPLACE PROCEDURE VALIDATION.VALID_LABEL_CONDITION(input varchar)
+    RETURNS BOOLEAN
+    LANGUAGE SQL
+AS
+BEGIN
+    let c varchar := (select parse_json(:input):condition);
+    let stmt varchar := (select 'select case when ' || :c || ' then true else false end from reporting.enriched_query_history limit 1');
+    execute immediate stmt;
+    return true;
```
rymurr, line 155: should this return the output of execute immediate?

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,122 @@ SELECT *,$$;
...
+    execute immediate stmt;
+    return false;
+EXCEPTION
+    WHEN STATEMENT_ERROR THEN
+        return true;
+    WHEN OTHER then
+        return false;
+END;
+
+create or replace function VALIDATION.wrap_simple_condition(condition text)
```
rymurr, line 175: should this and the other simple functions move into the 036 file?

rymurr (commented):
I like this! Looking goood. Can we get some accurate timings for this?

bootstrap/004_labels.sql
joshelser: I'm not sure we can use `execute immediate` in this case. `c` is just a string, but the `case` statement in the query needs to be a condition which evaluates to a boolean.

I can't think of any tricks to your suggestion work (I thought string concat to build the query is the only solution). Have I missed one?

bootstrap/004_labels.sql
joshelser: i think we rarely hit the else branch (often raise an exception), but yes that would be correct to do.

joshelser: oh, no, that's not right. We are just testing the compilation here. The actual row we get from query history may evaluate to true or false..

bootstrap/0036_sql_tools.sql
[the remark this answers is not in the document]
joshelser: Going to resolve as we have something very different than before which I think is much easier to comprehend.

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,154 @@
...
+    execute immediate stmt;
+    let rs resultset := (select false);
+    return table(rs);
+EXCEPTION
+    WHEN STATEMENT_ERROR THEN
+        let rs resultset := (select true);
+        return table(rs);
+    WHEN OTHER then
+        let rs resultset := (select false);
+        return table(rs);
```
joshelser, line 146: I had to change this proc and the previous one to returning a table so https://github.com/sundeck-io/OpsCenter/pull/591/files#diff-4d8db5ec382ac980083e32c616facb93597373fbc86135753e973a6fb38493a3R30 works w/o special handling.

I could capture the `boolean` and have a single `table(rs)` at the very end, but I really wish I didn't have to.

Short of wrapping a anonymous procedure around the real procedure, I couldn't come up with a way to "rename" the output column from the proc (the column name of the result is always the name of the procedure itself).

bootstrap/0036_sql_tools.sql
```diff
@@ -0,0 +1,154 @@
...
+
+CREATE OR REPLACE PROCEDURE VALIDATION.VALID_LABEL_CONDITION(input varchar)
+    RETURNS TABLE(result BOOLEAN)
+    LANGUAGE SQL
+AS
+BEGIN
+    let c varchar := (select parse_json(:input):condition);
+    let stmt varchar := (select 'select ' || :c || ' from reporting.enriched_query_history limit 1');
+    execute immediate :stmt;
+    -- if the condition is invalid, we let the exception propagate.
```
rymurr, line 123: why do we allow the exception to propagate here but below we capture it?

rymurr (dismissed):
1 minor question and seems tests are still failing. I think it looks great though.

Any comments on time and improvements?

joshelser:
> Any comments on time and improvements?

Yeesh, we took a step backwards for ungrouped label creation

`ADMIN.CREATE_LABEL()` is taking ~6s. Of that, the current breakdown is:
* `admin.validate` takes ~3.2s
* `admin.write` takes ~1.2s
* `internal.update_label_view` takes ~1.8s

The sum of queries executed by `admin.write` are around 700ms, so 1.2s seems long. The two "complex" validations represent ~50% of the total 3.2s of validation. I'll see if there's something obvious we can do differently here.

I'll re-evaluate with a hybrid table in a momen.t

joshelser:
> I'll re-evaluate with a hybrid table in a moment

Making `internal.labels` a hybrid table with `label_id primary key` drops the execution time of the `MERGE` statement inside `admin.write` from ~400ms to ~250ms.

bootstrap/0036_sql_tools.sql
joshelser: We don't managed the exceptions below so we can distinguish `statement_error` from `other` exceptions. In this procedure, any exception indicates a failed validation (and we can let the caller handle it since we have the caller doing this as a fail-safe).

bootstrap/004_labels.sql
```diff
@@ -114,38 +113,99 @@ SELECT *,$$;
...
+        -- group_name and group_rank must be null
+        (VALIDATION.IS_NULL('group_name') || ' and ' || VALIDATION.IS_NULL('group_rank'), '{"message": "Group rank may only be provided for grouped labels"}'),
+        -- label name must be unique among all names
+        (VALIDATION.LABEL_ABSENT(), '{"message": "A label with this name already exists", "create_only": true}'),
+        -- label name must be unique across all group_names
+        ('(select count(*) = 0 from internal.labels where group_name = f:name and group_name is not null)', '{"message": "A label group already exists with this name", "create_only": true}'),
+        -- Condition must compile
+        ('CALL VALIDATION.VALID_LABEL_CONDITION(?)', '{"message": "Invalid label condition", "complex": true}'),
+        -- make sure label name doesn't exist as query history column
+        ('CALL VALIDATION.IS_VALID_LABEL_NAME(?)', '{"message": "Label name cannot be the same as a column in REPORTING.ENRICHED_QUERY_HISTORY", "complex": true}')
```
joshelser, line 133: These two checks as standalone procedures (versus inline blocks) cost ~1s.

jacques-n:
I think we should be able to make all validation a single SQL select query. Prove me wrong :D

joshelser:
> I think we should be able to make all validation a single SQL select query. Prove me wrong :D

I just pushed the latest (which has grouped and ungrouped labels passing their tests). I did a quick prototype with just the simple validations in one `select` but ignored the "complex" validations (which we know for sure are the time-suck).

I'll see if I can get all of the checks into a single query. The one problem I can immediately think of is that we rely on an exception to be raised to verify that a label name doesn't duplicate a name in `enriched_query_history`. Will have to take a different approach on that check.

rymurr:
Maybe an unpopular opinion: should we give this up and just go back to V1 of these? V1 beign specialized procedures for each crud type? I seem to recall those being faster than this. My memory:

v1: First we had standalone hand coded procs for each entity type. Reasonably fast but hard to extend and had to be called as stored procs from streamlit
v2: Move into a generic python structure. Easy to extend and very fast from streamlit but slow from stored procs
v3: since we are getting rid of streamlit we don't care if its fast from streamlit. We need fast from stored procs. Do we need it to be easy to extend?

My suggestion, in the interest of expediency is to either defer this whole problem or to stop trying to make generic validation. Thoughts?

joshelser:
> Maybe an unpopular opinion: should we give this up and just go back to V1 of these? V1 beign specialized procedures for each crud type?

I had the same thought yesterday. I almost installed an old version just to confirm that they were actually faster before.

PR #593, opened 2024-03-21
Rename FLOW_NAME and ALT_WAREHOUSE_NAME

joshelser:
This is using the old columns. When the underlying schema is updated, I will update the view to use the renamed columns.


deploy/sundeck_sharing.sql
```diff
@@ -15,10 +15,12 @@ BEGIN
...
             SNOWFLAKE_QUERY_ID,
             SUNDECK_QUERY_ID,
-            FLOW_NAME,
+            FLOW_NAME, -- TODO drop from app-package view once released
+            FLOW_NAME as BROKER_NAME, -- TODO use BROKER_NAME once Sundeck is updated
             QUERY_TEXT_RECEIVED,
             QUERY_TEXT_FINAL,
-            ALT_WAREHOUSE_ROUTE,
+            ALT_WAREHOUSE_ROUTE, -- TODO drop from app-package view once released
+            ALT_WAREHOUSE_ROUTE as UPDATED_WAREHOUSE_TARGET, -- TODO use UPDATED_WAREHOUSE_TARGET once Sundeck is updated
```
joshelser, line 23: These changes may look strange. The view in the App Package will be used both by v2.6 (current, publicly released) of the native app and the new patch version which contains this commit.

We need to make sure that both the current version and previous version(s) continue to work (i.e. have the expected columns).

rymurr (approved):

PR #598, opened 2024-03-27
Fix pydantic version

rymurr:


app/ui/crud.zip
joshelser, line 1: Shouldn't have this checked in and it should be in `.gitignore` and isn't.

joshelser (commented):
I'm really surprised that pydantic 2.x showed up (assuming by pinning pydantic 1.x). I never thought that would happen because of the native code in pydantic2.

app/ui/crud.zip
rymurr: I already deleted it from `main` will also add to `.gitignore`

PR #599, opened 2024-03-27
Run this job once a day.

vqmarkman:
I _think_ this should be enough.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -6,6 +6,10 @@ on:
   push:
     branches: [main]
 
+  schedule:
+    # 5:33am UTC/9:33pm PST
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -6,6 +6,10 @@ on:
   push:
     branches: [main]
 
+  schedule:
```
rymurr, line 9: I think we should run just e2e tests and we should run it on the 'real' deployed opscenter version

vqmarkman: well, that's what I suggested this morning. Let's talk before I do anything else. Thanks!

rymurr: then we both agree, lets do it!

PR #600, opened 2024-03-28
Nightly e2e cypress tests.

vqmarkman:
Decided to temporarily add a separate nightly run of cypress tests off the latest main branch with devdeploy.py (and not deploy.py)

Successful run of this workflow is [here](https://github.com/vqmarkman/OpsCenter/actions/runs/8462980728/job/23185154857) 

The reason I went with devdeploy.py: 

After I successfully deployed the app in my account with deploy.py (semi real deployment) I realized  (a)  native app very slow to start up and load pages (b) I will need to re-implement login to Snowflake in order to start the app as it can only be started from the UI :( ...

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,69 @@
+name: Nightly OpsCenter Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    # 5:33am UTC/9:33pm PST
+    - cron: '33 5 * * *'
+
+run-name: Nightly Opcenter Tests
+
+concurrency:
+  group: ci-post-deployment-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  cypress-nightly-tests:
+    runs-on: ubuntu-latest-4-cores
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,69 @@
+name: Nightly OpsCenter Tests
+
+on:
+  workflow_dispatch:
```
vqmarkman, line 4: Not sure we should allow general public to start this job on demand. I remember discussing this with Josh many month ago and he did not want to do it. Please let me know if we should remove this ability.

rymurr:
I don't think `devdeploy` is useful here. We will likely have to figure out somethign for the snowflake login bit. The errors we've been experiencing yesterday were primarily with things that would only be noticed when we deploy streamlit to Snowflake. We would not have seen some of the errors just running local streamlit like we do in devdeploy

vqmarkman:
> We will likely have to figure out somethign for the snowflake login bit
I had it working at some point (pretty crude ugly hack, but it did work). Will  resurrect it.

PR #601, opened 2024-03-28
add proprietary code and have it get installed

rymurr:


joshelser (approved):
lgtm. Only concern is when we get into stable-state, making sure we avoid accidentally running deploy.py w/o having the extra logic in the local checkout, but we don't need to solve that today.

rymurr:
> Only concern is when we get into stable-state, making sure we avoid accidentally running deploy.py w/o having the extra logic in the local checkout, but we don't need to solve that today.

Also concerned...my thinking was that we will have to do some automation (in another repo) to do the deploy correctly

joshelser:
> my thinking was that we will have to do some automation (in another repo) to do the deploy correctly

Yep, it's probably time we deploy++ ourselves to reduce our cognitive load for promotions.

PR #602, opened 2024-03-28
update workflows

rymurr:


joshelser (commented):
was `submodules: recursive` not doing the correct thing?

rymurr:
it was doing the right thing but this to prevent trying to pull the new submodule from #601

joshelser:
Ah! Thanks, now I understand.

PR #603, opened 2024-03-28
New tables for reporting

rymurr:


bootstrap/023_simple_materializations.sql
```diff
@@ -0,0 +1,95 @@
...
+    let vw varchar := 'REPORTING.' || :table_name;
+    create or replace view identifier(:vw) as select * from identifier(:tbl);
+end;
+
+create table if not exists internal_reporting_mv.serverless_task_history as select * from account_usage.serverless_task_history where 1=0;
+create table if not exists internal_reporting_mv.task_history as select * from account_usage.task_history where 1=0;
+create table if not exists internal_reporting_mv.sessions as select * from account_usage.sessions where 1=0;
+create or replace view reporting.serverless_task_history as select * from account_usage.serverless_task_history where 1=0;
+create or replace view reporting.task_history as select * from account_usage.task_history where 1=0;
+create or replace view reporting.sessions as select * from account_usage.sessions where 1=0;
```
joshelser, line 87: You don't want the `where 1=0` on the view def'n, do you?

joshelser (changes requested):
I double-checked your view def'n against the snowflake view. Other than (what looks to me like) the inadvertent copy-paste on the reporting view creations, LGTM.

bootstrap/023_simple_materializations.sql
rymurr: good spot, fixed

PR #604, opened 2024-03-28
remove run_as_app

rymurr:
Some changes have to propagate to prod on the SaaS before this can be merged


app/ui/pages/10_Settings.py
```diff
@@ -20,7 +22,7 @@ def invalid_number(string: str) -> Optional[str]:
 
 
 def get_task_status(dbname):
-    sql = f"""call admin.run_as_app('describe task TASKS."{dbname}";');"""
+    sql = f"""describe task TASKS."{dbname}";"""
```
joshelser, line 25: I'm surprised that we had this wrapped in the `run_as_app`. Do you remember if there was a reason we had this one?

joshelser (dismissed):

app/ui/pages/10_Settings.py
rymurr: I don't think tasks were exposed to admin in early iterations

app/ui/pages/10_Settings.py
```diff
@@ -132,17 +134,28 @@ def task_listing(
...
-                alter task TASKS.SFUSER_MAINTENANCE {get_task_state(pm)};
-                alter task TASKS.USER_LIMITS_MAINTENANCE {get_task_state(user_limits)};
-            end;
-            $$);
-            """
-            connection.execute(sql)
+            with connection.Connection.get() as conn:
+                with snowpark_session(conn) as session:
+                    t = Task(task_name="WAREHOUSE_EVENTS_MAINTENANCE")
+                    if wem:
```
jacques-n, line 140: Can this be factored to reduce duplication?

Especially to avoid redefining `t` to mean different things.

What you really have is a set of `{target_state: boolean, task_name: string}` objects that you're apply the same operation to.

rymurr: yup, I  was lazy. Fixed

joshelser (approved):
RIP run_as_admin

PR #606, opened 2024-03-28
warehouse load history

rymurr:


rymurr:
calling  `call internal.refresh_one_warehouse_load_history('COMPUTE_WH');`

produces the sql statement
``` sql
insert into internal_reporting_mv.warehouse_load_history select start_time, end_time, warehouse_name, avg_running, avg_queued_load, avg_queued_provisioning, avg_blocked from table(dryan.information_schema.warehouse_load_history(date_range_start => to_timestamp_ltz('2024-03-14 10:43:51.973'), date_range_end => to_timestamp_ltz('2024-03-14 18:43:51.973'), warehouse_name => 'COMPUTE_WH'))
```

but fails with
```
Uncaught exception of type 'STATEMENT_ERROR' on line 19 at position 8 : Stored procedure execution error: Requested information on the current user is not accessible in stored procedure.
```

bootstrap/023_simple_materializations.sql
```diff
@@ -93,3 +94,39 @@ begin
...
+        let oldest_running timestamp := 0::timestamp;
+
+        if (input is not null) then
+            oldest_running := coalesce(input:oldest_running::timestamp, 0::timestamp);
+        end if;
+
+        let stmt varchar := 'insert into internal_reporting_mv.warehouse_load_history select start_time, end_time, warehouse_name, avg_running, avg_queued_load, avg_queued_provisioning, avg_blocked from table(information_schema.warehouse_load_history(date_range_start => to_timestamp_ltz(\'{start_time}\'), date_range_end => to_timestamp_ltz(\'{end_time}\'), warehouse_name => \'{warehouse_name}\'))';
+        let start_time timestamp := (select greatest(:oldest_running, dateadd(day, -14, current_timestamp())));
+        let res resultset := (select tools.templatejs(:stmt,
+            object_construct('start_time', dateadd(hours, value, :start_time)::text, 'end_time', dateadd(hours, value+8, :start_time)::text, 'warehouse_name', :warehouse_name))
```
jacques-n, line 125: in general, I'd be inclined to use object literal format instead of object_construct. is clearer to read. Does it not work in this context? (e.g. with use of :warehouse_name, :start_time)

e.g. 
```
{
  'start_time': (dateadd(hours, value, :start_time)::text), 
  'end_time': (dateadd(hours, value+8, :start_time)::text)
} 
```

jacques-n: Tested, seems like it should work. 

BEGIN
let w := 'baz';
let rs resultset  := (select {'foo': query_text, 'bar': :w} from opscenter.reporting.enriched_query_history limit 10);
return table(rs);
end;

rymurr: fixed!

bootstrap/023_simple_materializations.sql
```diff
@@ -93,3 +94,42 @@ begin
...
+        if (input is not null) then
+            oldest_running := coalesce(input:oldest_running::timestamp, 0::timestamp);
+        end if;
+
+        let stmt varchar := 'insert into internal_reporting_mv.warehouse_load_history select start_time, end_time, warehouse_name, avg_running, avg_queued_load, avg_queued_provisioning, avg_blocked from table(information_schema.warehouse_load_history(date_range_start => to_timestamp_ltz(\'{start_time}\'), date_range_end => to_timestamp_ltz(\'{end_time}\'), warehouse_name => \'{warehouse_name}\'))';
+        let start_time timestamp := (select greatest(:oldest_running, dateadd(day, -14, current_timestamp())));
+        let res resultset := (select tools.templatejs(:stmt,
+            {
+              'start_time': dateadd(hours, value, :start_time)::text,
+              'end_time': dateadd(hours, value+8, :start_time)::text
```
joshelser, line 127: need `warehouse_name` or does that somehow get substituted in?

bootstrap/023_simple_materializations.sql
```diff
@@ -93,3 +94,42 @@ begin
...
+        end_time timestamp_ltz,
+        warehouse_name text,
+        avg_running NUMBER(38,2),
+        avg_queued_load NUMBER(38,2),
+        avg_queued_provisioning NUMBER(38,2),
+        avg_blocked NUMBER(38,2));
+create or replace view reporting.warehouse_load_history as select * from internal_reporting_mv.warehouse_load_history;
+
+
+CREATE OR REPLACE PROCEDURE internal.refresh_one_warehouse_load_history(warehouse_name varchar) RETURNS table(varchar) LANGUAGE SQL
```
joshelser, line 108: I think you are accessing the varchar column from `RETURNS table(varchar)` as `sql` in `TASKS.WAREHOUSE_LOAD_MAINTENANCE`. Should this be `RETURNS table(sql varchar)` to be a little more obvious?  (tbh, it's not clear to me how this works to not give a column name when returning a table from a proc, but am assuming snowflake figures this out :D)

bootstrap/090_post_setup.sql
```diff
@@ -286,6 +286,39 @@ exception
...
+    for wh_row in wh_cur do
+        let wh_name varchar := wh_row.name;
+        let input variant := (select output from INTERNAL.TASK_WAREHOUSE_LOAD_EVENTS where success and warehouse_name = :wh_name order by run desc limit 1);
+        begin
+            let stmt resultset := (call internal.refresh_one_warehouse_load_history(:wh_name));
+            let stmt_cur cursor for stmt;
+            let counter number := 0;
+            for stmt_row in stmt_cur do
+                execute immediate stmt_row.sql;
+                let new_count number := (select * from TABLE(RESULT_SCAN(LAST_QUERY_ID())));
```
joshelser, line 307: we're collecting the number of rows inserted into our materialized table from the table function? A comment would help make this clear (through the ~3 layers of indirection) or maybe just calling this somethign like `rows_inserted_for_wh` instead of the generic `count`

joshelser (changes requested):
Looks real close to me.

bootstrap/023_simple_materializations.sql
rymurr: good shout

bootstrap/023_simple_materializations.sql
rymurr: weird, I put that somewhere but must have accidentally dropped it. Fixed

bootstrap/090_post_setup.sql
rymurr: fixed

joshelser (approved):

joshelser (approved):

PR #608, opened 2024-04-01
Always re-create external functions during UPGRADE_CHECK()

joshelser:
The linking to Sundeck path will set a new definition for internal.get_ef_token() but we don't want to proactively re-create the external functions because this takes a few seconds.

Instead, recreate the functions in UPGRADE_CHECK rather than in FINALIZE_SETUP which UPGRADE_CHECK may or may-not call.

### DIFF
--- a/bootstrap/092_service_account_setup.sql
+++ b/bootstrap/092_service_account_setup.sql
@@ -40,6 +40,10 @@ returns varchar
 language sql
 as
 begin
+    -- Always call setup_external_functions to ensure the external_functions are re-created with the latest opscenter
+    -- token as Sundeck likely invalidated the previous token.
+    CALL admin.setup_external_functions('opscenter_api_integration');
+
     let version varchar;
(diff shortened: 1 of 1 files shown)


rymurr (approved):

bootstrap/092_service_account_setup.sql
```diff
@@ -40,6 +40,10 @@ returns varchar
 language sql
 as
 begin
+    -- Always call setup_external_functions to ensure the external_functions are re-created with the latest opscenter
+    -- token as Sundeck likely invalidated the previous token.
+    CALL admin.setup_external_functions('opscenter_api_integration');
```
rymurr, line 45: why call this here? I don't understand what is special about this proc? Maybe the name is just thorwing me off

joshelser:
Discussed with Ryan and Vicky, IRL. Going to solve this in a different way.

PR #609, opened 2024-04-02
Add metering history to persisted tables

rymurr:
Schema as per https://docs.snowflake.com/en/sql-reference/account-usage/warehouse_metering_history

note also the data isn't stable for 6 hours...


joshelser (approved):
Looks like there is a new column `READER_ACCOUNT_NAME` in warehouse_metering_history since we created the dummy views in in 003..sql. Probably a good idea to update that while we're here.

rymurr:
> Looks like there is a new column READER_ACCOUNT_NAME in warehouse_metering_history

from the docs "Column only included in view in READER_ACCOUNT_USAGE schema."

PR #614, opened 2024-04-04
Move admin.update_settings to SQL

joshelser:
I know you said this morning that you didn't want to go down this route, but I'm worried about the variability in latency impacting other things.

### DIFF
--- a/bootstrap/050_settings.sql
+++ b/bootstrap/050_settings.sql
@@ -39,7 +39,7 @@ BEGIN
     end if;
 
 	call internal.set_config(:name, :value) into :res;
-	return res;
+	return "";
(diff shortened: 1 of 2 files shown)


rymurr (commented):
So we have lost validation on these settings?

rymurr:
ive added the validations back in

rymurr:

joshelser:
> So we have lost validation on these settings?

Yeah, I thought the likelihood of someone actually setting these by hand was low. I mostly thought that because i hadn't thought of just calling CONVERT_TIMEZONE like you did to validate the provided TZ. Nice trick.

PR #615, opened 2024-04-04
Fix validity check

rymurr:


rymurr:

joshelser (commented):
👍🏼

PR #619, opened 2024-04-05
Create external functions async in finalize_setup_with_servcie_account

joshelser:
It feels insurmountable to creates these synchronously which is frustrating as it (non-scientifically) seems to add 3-5 seconds.


joshelser:
Defer this change for now.

PR #627, opened 2024-04-09
Cleanup for latency's sake

joshelser:
* Remove internal.get_ef_url() and use the row from the internal.config table
* Remove unused internal.get_tenant_url UDF
* Replace the 3x `internal.set_config()` calls in `admin.finalize_setup_with_service_account() with a single `merge`
* Remove the `internal.reference_management` table as we do nothing with it.
* Create the normal task log table and view for `UPGRADE_CHECK`

This PR intentionally does not move code around for additional parallelism as we discussed. That will come in a second PR after verifying these changes work as expected.

Will be doing some manual poking and prodding on this PR, too.

### DIFF
--- a/bootstrap/021_external_functions.sql
+++ b/bootstrap/021_external_functions.sql
@@ -320,12 +320,6 @@ $$
 $$;
 
 
-create function if not exists internal.get_tenant_url()
-    returns string
-    language javascript
-    as
-    'throw "You must configure a Sundeck token to use this.";';
-
 create function if not exists internal.get_ef_token()
     returns string
     language javascript
--- a/bootstrap/021_external_functions.sql
+++ b/bootstrap/021_external_functions.sql
@@ -513,3 +513,9 @@ BEGIN
     -- Success!
     return object_construct();
 END;
+
+-- remove deprecated functions if they still exist.
+drop function if exists internal.get_ef_url();
+drop function if exists internal.get_tenant_url();
+drop function if exists internal.setup_ef_url(string);
+drop function if exists internal.setup_sundeck_token(string, string);
--- a/bootstrap/021_external_functions.sql
+++ b/bootstrap/021_external_functions.sql
@@ -320,12 +320,6 @@ $$
 $$;
 
 
-create function if not exists internal.get_ef_url()
-    returns string
-    language javascript
-    as
-    'throw "You must configure a Sundeck token to use this.";';
-
 create function if not exists internal.get_tenant_url()
     returns string
     language javascript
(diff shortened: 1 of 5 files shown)


bootstrap/021_external_functions.sql
```diff
@@ -320,35 +320,12 @@ $$
...
-    as
-    'throw "You must configure a Sundeck token to use this.";';
-
 create function if not exists internal.get_ef_token()
     returns string
     language javascript
     as
     'throw "You must configure a Sundeck token to use this.";';
 
-create or replace procedure internal.setup_ef_url(url string) RETURNS STRING LANGUAGE SQL AS
```
rymurr, line 341: should we add  a `DROP IF EXISTS`?

bootstrap/090_post_setup.sql
```diff
@@ -1,5 +1,5 @@
 
-create or replace procedure admin.finalize_setup()
+create or replace procedure admin.finalize_setup(log_output boolean default true)
```
rymurr, line 2: why would this be optional?

bootstrap/092_service_account_setup.sql
```diff
@@ -18,19 +15,26 @@ begin
         CALL ADMIN.UPGRADE_CHECK();
     grant MONITOR, OPERATE on TASK TASKS.UPGRADE_CHECK to APPLICATION ROLE ADMIN;
 
-    call internal.set_config('tenant_url', :web_url);
-    call internal.set_config('url', :url);
-
-    let tenant_id text := (select split_part(:web_url, '/', -1));
-    call internal.set_config('tenant_id', :tenant_id);
+    -- Merge all three properties into the config table in one merge statement.
+    MERGE INTO internal.config AS target
```
rymurr, line 19: I had  a slightly different syntax (some extra brackets I think) did you test this?

bootstrap/021_external_functions.sql
joshelser: yep, definitely.

bootstrap/092_service_account_setup.sql
joshelser: Yeah, I called the procs locally and it's working. What'd you have in mind?

Looking at this again, I should also take the opportunity to explicitly pass the tenant ID rather than parsing it out.

bootstrap/090_post_setup.sql
joshelser: Ya got me: this is the one part of this PR I wasn't happy with.

I added logging to `admin.upgrade_check` to later capture any errors in that procedure that happen outside of `admin.finalize_setup()`

However, I also didn't want to break someone being able to `call admin.finalize_setup()` nor miss out on the logging if they do call that procedure. However, we need to have some logging in `admin.upgrade_check` as we are doing more than just calling `admin.finalize_setup()`.

Finally, I also didn't want to have two tables to generate logging into.

The "best" solution I came up with was to have `admin.upgrade_check()` call `admin.finalize_setup(false)` to prevent both procedures from generating a log message. I still think it's ugly.

rymurr: I don't understand the problem you are trying to solve w/ the boolean? Why cant both log?

bootstrap/092_service_account_setup.sql
rymurr: > Yeah, I called the procs locally and it's working. What'd you have in mind?

all good then. Prob a non-event. Just wanted to check that there wasn't something funny going on

bootstrap/090_post_setup.sql
joshelser: I wanted to avoid the upgrade_check task running every 24 hours and generating two rows of output into the same table. If you think this isn't worth the code to maintain it, i'll just drop it.

rymurr: yeah, I don't care at all...

rymurr (approved):

PR #628, opened 2024-04-09
New procedures which break apart finalize_setup_from_service_account()

joshelser:
We must always call setup_external_functions in upgrade_check.

The three pieces we can now call concurrently (in place of `finalize_setup_with_service_account()`) are:
* `admin.update_reference(text, text, text)`
* `admin.create_upgrade_check()`
* `admin.update_sundeck_configuration(text, text, text)`

### DIFF
--- a/bootstrap/039_reference_management.sql
+++ b/bootstrap/039_reference_management.sql
@@ -3,8 +3,8 @@ drop view if exists catalog.references;
 drop table if exists internal.reference_management;
 
 create or replace procedure admin.update_reference(ref_name string, operation string, ref_or_alias string)
- language sql
  returns string
+ language sql
  as
 begin
   SYSTEM$LOG_INFO('Updating reference: ' || ref_name || ' operation: ' || operation || ' ref_or_alias: ' || ref_or_alias);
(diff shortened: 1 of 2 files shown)


bootstrap/039_reference_management.sql
```diff
@@ -25,6 +26,5 @@ begin
     else
        return 'Unknown operation: ' || operation;
   end case;
-  return 'Success';
```
rymurr, line 28: this lin elooks weird. Why are we getting rid of information?

rymurr (dismissed):

bootstrap/039_reference_management.sql
joshelser: We have a mixed bag of when it comes to what does it mean when a procedure returns a string. Sometimes it has a sentinel string like "Success" and sometimes it is an error message (when we detect something specific was wrong).

My goal was to make this procedure consistent with the other ones in this PR so we have fewer special cases from the caller's POV. If you'd rather not continue that convention, that's also fine (raising exceptions is probably what we should have been doing all along).

rymurr (approved):

joshelser:
One minor conflict on rebase

rymurr (approved):

PR #637, opened 2024-04-16
forgot to create reporting views

rymurr:


joshelser (dismissed):
I didn't double-check that the list of objects is correct, but these 3 additions look good.

joshelser (approved):

PR #643, opened 2024-04-24
Create a view that reports on materialization status of tables

joshelser:
~~* Starts with QUERY_HISTORY and WAREHOUSE_EVENTS_HISTORY~~
~~* Doesn't try to look at materializations that are running right now (e.g. those that need information_schema.task_history or an update to how we update the task tables).~~
~~* Added some basic unit tests before doing more.~~

Substantial changes from yesterday:
* ~Still only reports on tables that are materialized by query history maintenance and warehouse events maintenance tasks~ Reports on all tables in REPORTING (sans the HOURLY, DAILY variants)
* Uses `SYSTEM$TASK_RUNTIME_INFO` to get a graph_run_id that lets us join against the snowflake `TASK_HISTORY` to look up the `QUERY_ID` for each materialization
* Looks at both the TASK_HISTORY table function (for the "next" and recently completed tasks) and the materialized task_history (for historical tasks).
* Joins against QUERY_HISTORY on query_id to add warehouse_size
* Addresses some feedback from Jacques on schema
* * Each row reports the last full, last incremental, and next materialization for the user-facing views
* * Drops the explicit ranges of data included in tables.

~~I've only been testing in devdeploy. Need to test in a standalone app today.~~
~~Then, once we're happy with this approach, add in the rest of the reporting tables.~~

Have tested in both devdeploy. standalone app deploys, but i need to dbl check with fresh eyes.


bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,46 @@
+
+create or replace view internal.all_task_history(run, success, input, output, table_name) as
+    select run, success, input, output, 'QUERY_HISTORY' from internal.task_query_history
+    union all
+    select run, success, input, output, 'WAREHOUSE_EVENTS_HISTORY'  from internal.task_warehouse_events;
```
joshelser, line 5: I didn't want to explicitly create this but I was struggling with a way to compose this from the `task_tables` CTE below.

I feel like I should be able to use a static list of task history tables (internal.task_query_history, internal.task_warehouse_events) and compose an equivalent `select` statement without dynamically creating sql.

Any thoughts?

joshelser: Solution came to me once I stopped looking at the computer -- just push this query into the CTE and join on 'table_name'. 🤦🏼‍♀️

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,100 @@
...
+        union all
+        select run::TIMESTAMP_LTZ, success, input, output, 'WAREHOUSE_EVENTS_HISTORY', 'WAREHOUSE_EVENTS_MAINTENANCE' from internal.task_warehouse_events
+        union all
+        select run::TIMESTAMP_LTZ, success, input, output, table_name, 'SIMPLE_DATA_EVENTS_MAINTENANCE' from internal.task_simple_data_events
+    )
+), sf_task_history as (
+    select name as task_name, query_id, graph_run_group_id, state, scheduled_time, QUERY_START_TIME,
+    from table(information_schema.task_history(SCHEDULED_TIME_RANGE_START => TIMESTAMPADD(HOUR, -3, current_timestamp()), result_limit => 1000))
+    WHERE database_name in (select current_database()) and schema_name = 'TASKS'
+    union
```
joshelser, line 43: Creating the view inside the setup script fails:

```
File: 051_materialization_status.sql
[Uncaught exception of type 'STATEMENT_ERROR' on line 21 at position 0 : Stored procedure execution error: Requested information on the current user is not accessible in stored procedure.] was primary error.
```

I'm guessing because of the setup script limitations rather than a native app limitation. https://docs.snowflake.com/en/developer-guide/native-apps/adding-application-logic#limitations-when-calling-context-functions-from-the-setup-script

joshelser: Jacques to the rescue -- have to create it from within a task in the native app. Same sort of thing that probe_monitoring is doing.

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,158 @@
...
+            select run::TIMESTAMP_LTZ, success, input, output, 'QUERY_HISTORY', 'QUERY_HISTORY_MAINTENANCE'   from internal.task_query_history
+            union all
+            select run::TIMESTAMP_LTZ, success, input, output, 'WAREHOUSE_EVENTS_HISTORY', 'WAREHOUSE_EVENTS_MAINTENANCE' from internal.task_warehouse_events
+            union all
+            select run::TIMESTAMP_LTZ, success, input, output, table_name, 'SIMPLE_DATA_EVENTS_MAINTENANCE' from internal.task_simple_data_events
+        )
+    ), sf_task_history as (
+        -- Our TASK_HISTORY view is updated with a 3hour delay. Limit how far we back we look in the UDTF.
+        select name as task_name, query_id, graph_run_group_id, state, scheduled_time, QUERY_START_TIME,
+        from table(information_schema.task_history(SCHEDULED_TIME_RANGE_START => TIMESTAMPADD(HOUR, -3, current_timestamp()), result_limit => 10000))
```
joshelser, line 59: Having to set `result_limit =10000` isn't great here (and the other UDTF usages) because it's slow and not optimized. However, because we (generally) have a delay of 180minutes for each of the ACCOUNT_USAGE views, we don't have a great solution other than pay this cost.

Because the QH/WEH tasks are running every 1hour, we won't be able to join the query_id/warehouse_size unless we use the UDTF.  At that point, it's just a matter of luck as to how active a user's snowflake account it.

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,158 @@
...
+        select * from sf_task_history where internal.task_pending(state)
+    ), fullmat as (
+        SELECT th.table_name, th.task_name, run, output['end']::TIMESTAMP_LTZ as end, success,
+            output['SQLERRM'] as error_message, ct.query_id,
+            COALESCE(qht.warehouse_size, qh.warehouse_size) as warehouse_size,
+        FROM task_history th
+        LEFT JOIN completed_tasks ct ON th.output['task_run_id'] = ct.GRAPH_RUN_GROUP_ID
+        LEFT JOIN internal_reporting_mv.task_history th
+        -- Get the warehouse_size from QueryHistory. Check both the view and UDTF to avoid gaps.
+        LEFT JOIN internal_reporting_mv.query_history_complete_and_daily_incomplete qh ON ct.query_id = qh.query_id
```
joshelser, line 77: Pushing this into a join (rather than another CTE branch as task_history above) helped with performance, I think (unless I just deluded myself with a small number of rows materialized). If this is slow, I suspect we could cluster_by `query_id`?

bootstrap/090_post_setup.sql
```diff
@@ -329,22 +330,39 @@ BEGIN
...
+-- The admin.materialization_status view has to be created from a task because it reads from the TASK_HISTORY UDTF.
+CREATE OR REPLACE TASK TASKS.CREATE_MATERIALIZATION_STATUS
+    ALLOW_OVERLAPPING_EXECUTION = FALSE
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "X-Small"
+    AS
+DECLARE
+  sql STRING;
+BEGIN
+    call internal.create_materialization_status() into :sql;
+    execute immediate sql;
```
joshelser, line 358: Have to play this trick (like for probe_monitoring), else creating a view or UDTF that interacts with the snowflake.information_schema UDTFs will fail.

bootstrap/051_materialization_status.sql
joshelser: > Having to set result_limit =10000 isn't great here (and the other UDTF usages) because it's slow and not optimized. However, because we (generally) have a delay of 180minutes for each of the ACCOUNT_USAGE views, we don't have a great solution other than pay this cost.

The metadata operation for the query_history UDTF with a result_limit of 10K takes 10s on its own. That's a non-starter.

bootstrap/051_materialization_status.sql
joshelser: wrong table name here, but perf is not an issue, partitions are already getting pruned (e.g. scanned 12/4xx)

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,159 @@
...
+        union
+        select name as task_name, query_id, graph_run_group_id, state, scheduled_time, QUERY_START_TIME,
+        from reporting.task_history
+        WHERE database_name in (select current_database()) and schema_name = 'TASKS'
+    ), completed_tasks as (
+        select * from sf_task_history where internal.task_complete(state)
+    ), pending_tasks as (
+        select * from sf_task_history where internal.task_pending(state)
+    ), fullmat as (
+        SELECT th.table_name, th.task_name, run, output['end']::TIMESTAMP_LTZ as end, success,
```
joshelser, line 70: Thinking more, the intent was likely that we're reporting the range of data materialized, not the duration of the materialization task (user doesn't care about that)

bootstrap/051_materialization_status.sql
joshelser: Thought: can we push the task_history and query_history UDTF lookups into the respective tasks and emit the query_id and warehouse_size via `output` in the task log tables?

PR #644, opened 2024-04-27
Create a singular INTERNAL.TASK_LOG table which records query_id

joshelser:
Suggestion from Jacques on Friday after we discussed the slow performance of #643 (due to the task_history and query_history UDTFs).

* Separate what a task does from the metadata recording that we want to track
* Record the query_id for a task when it runs into that metadata
* Migrate the separate task log tables into one (with even context to filter the update of a specific object)

### DIFF
--- a/bootstrap/012_task_log.sql
+++ b/bootstrap/012_task_log.sql
@@ -38,6 +38,25 @@ BEGIN
         WHERE task_start = :start_time AND task_run_id = :task_run_id AND task_name = :task_name AND object_name = :object_name;
 END;
 
+-- Special case for the warehouse events task, which inserts extra rows into task log breaking out warehouse and cluster sessions details to
+-- ease reporting on each dataset.
+create or replace procedure internal.finish_warehouse_events_task(task_name text, start_time timestamp_ltz, task_run_id text, query_id text, input object, output object)
+    returns text
+    language sql
+AS
+BEGIN
+    let success boolean := (select :output['SQLERRM'] is null);
+    let cluster_range_min timestamp_ltz := (select :output['cluster_range_min']);
+    let cluster_range_max timestamp_ltz := (select :output['cluster_range_max']);
+    INSERT INTO INTERNAL.TASK_LOG(task_start, success, task_name, object_name, input, output, task_finish, task_run_id, query_id, range_min, range_max)
+    select :start_time, :success, :task_name, 'CLUSTER_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :cluster_range_min, :cluster_range_max;
+
+    let warehouse_range_min timestamp_ltz := (select :output['warehouse_range_min']);
+    let warehouse_range_max timestamp_ltz := (select :output['warehouse_range_max']);
+    INSERT INTO INTERNAL.TASK_LOG(task_start, success, task_name, object_name, input, output, task_finish, task_run_id, query_id, range_min, range_max)
+    select :start_time, :success, :task_name, 'WAREHOUSE_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :warehouse_range_min, :warehouse_range_max;
+END;
+
 -- Generic table that tasks should record their execution into.
 CREATE TABLE INTERNAL.TASK_LOG IF NOT EXISTS (task_start timestamp_ltz, success boolean, task_name varchar, object_name varchar,
     input variant, output variant, task_finish timestamp_ltz, task_run_id text, query_id text, range_min timestamp_ltz, range_max timestamp_ltz);
--- a/bootstrap/012_task_log.sql
+++ b/bootstrap/012_task_log.sql
@@ -14,18 +14,18 @@ $$
 $$;
 
 -- Inserts a row into TASK_LOG with the time the task started, the name of the object being materialized, the graph run ID and query_id for the task.
-create or replace procedure internal.start_task(task_name text, object_name text, start_time timestamp_ltz, task_run_id text, query_id text)
+create or replace procedure internal.start_task(task_name text, object_name text, start_time text, task_run_id text, query_id text)
     returns object
     language sql
 AS
 BEGIN
     let input object := (select output from INTERNAL.TASK_LOG where success AND task_name = :task_name AND object_name = :object_name order by task_start desc limit 1);
-    INSERT INTO INTERNAL.TASK_LOG(task_start, task_run_id, query_id, input, task_name, object_name) select :start_time, :task_run_id, :query_id, :input, :task_name, :object_name;
+    INSERT INTO INTERNAL.TASK_LOG(task_start, task_run_id, query_id, input, task_name, object_name) select :start_time::TIMESTAMP_LTZ, :task_run_id, :query_id, :input, :task_name, :object_name;
     return input;
 END;
 
 -- Updates the row created by START_TASK with the time the task finish, the success and output object of the task, and the min/max date range of the data that is now materialized.
-create or replace procedure internal.finish_task(task_name text, object_name text, start_time timestamp_ltz, task_run_id text, output object)
+create or replace procedure internal.finish_task(task_name text, object_name text, start_time text, task_run_id text, output object)
     returns text
     language sql
 AS
@@ -35,12 +35,12 @@ BEGIN
     let range_max timestamp_ltz := (select :output['range_max']);
     UPDATE INTERNAL.TASK_LOG
         SET success = :success, output = :output, task_finish = current_timestamp(), range_min = :range_min, range_max = :range_max
-        WHERE task_start = :start_time AND task_run_id = :task_run_id AND task_name = :task_name AND object_name = :object_name;
+        WHERE task_start = :start_time::TIMESTAMP_LTZ AND task_run_id = :task_run_id AND task_name = :task_name AND object_name = :object_name;
 END;
 
 -- Special case for the warehouse events task, which inserts extra rows into task log breaking out warehouse and cluster sessions details to
 -- ease reporting on each dataset.
-create or replace procedure internal.finish_warehouse_events_task(task_name text, start_time timestamp_ltz, task_run_id text, query_id text, input object, output object)
+create or replace procedure internal.finish_warehouse_events_task(task_name text, start_time text, task_run_id text, query_id text, input object, output object)
     returns text
     language sql
 AS
@@ -49,12 +49,12 @@ BEGIN
     let cluster_range_min timestamp_ltz := (select :output['cluster_range_min']);
     let cluster_range_max timestamp_ltz := (select :output['cluster_range_max']);
     INSERT INTO INTERNAL.TASK_LOG(task_start, success, task_name, object_name, input, output, task_finish, task_run_id, query_id, range_min, range_max)
-    select :start_time, :success, :task_name, 'CLUSTER_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :cluster_range_min, :cluster_range_max;
+    select :start_time::TIMESTAMP_LTZ, :success, :task_name, 'CLUSTER_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :cluster_range_min, :cluster_range_max;
 
     let warehouse_range_min timestamp_ltz := (select :output['warehouse_range_min']);
     let warehouse_range_max timestamp_ltz := (select :output['warehouse_range_max']);
     INSERT INTO INTERNAL.TASK_LOG(task_start, success, task_name, object_name, input, output, task_finish, task_run_id, query_id, range_min, range_max)
-    select :start_time, :success, :task_name, 'WAREHOUSE_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :warehouse_range_min, :warehouse_range_max;
+    select :start_time::TIMESTAMP_LTZ, :success, :task_name, 'WAREHOUSE_SESSIONS', :input, :output, current_timestamp(), :task_run_id, :query_id, :warehouse_range_min, :warehouse_range_max;
 END;
 
 -- Generic table that tasks should record their execution into.
--- a/bootstrap/012_task_log.sql
+++ b/bootstrap/012_task_log.sql
@@ -72,12 +72,12 @@ BEGIN
         BEGIN TRANSACTION;
         let config_key := (select 'MIGRATION_TASK_QUERY_HISTORY');
         let config_value text;
-        call internal.get_config(config_key) into :config_value;
+        call internal.get_config(:config_key) into :config_value;
         if (config_value is null) then
             INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
                 SELECT tqh.run, tqh.success, 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', tqh.input, tqh.output, null
                 FROM INTERNAL.TASK_QUERY_HISTORY tqh;
-            call internal.set_config(config_key, 'true');
+            call internal.set_config(:config_key, 'true');
         END IF;
         commit;
     end if;
@@ -86,12 +86,12 @@ BEGIN
         BEGIN TRANSACTION;
         let config_key := (select 'MIGRATION_TASK_WAREHOUSE_EVENTS');
         let config_value text;
-        call internal.get_config(config_key) into :config_value;
+        call internal.get_config(:config_key) into :config_value;
         if (config_value is null) then
             INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
                 SELECT twe.run, twe.success, 'WAREHOUSE_EVENTS_MAINTENANCE', 'WAREHOUSE_EVENTS_HISTORY', twe.input, twe.output, null
                 FROM INTERNAL.TASK_WAREHOUSE_EVENTS twe;
-            call internal.set_config(config_key, 'true');
+            call internal.set_config(:config_key, 'true');
         END IF;
         commit;
     end if;
@@ -100,13 +100,28 @@ BEGIN
         BEGIN TRANSACTION;
         let config_key := (select 'MIGRATION_TASK_SIMPLE_DATA_EVENTS');
         let config_value text;
-        call internal.get_config(config_key) into :config_value;
+        call internal.get_config(:config_key) into :config_value;
         if (config_value is null) then
             -- Different from the previous, carrying table_name into task_log as object_name
             INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
                 SELECT tsde.run, tsde.success, 'SIMPLE_DATA_EVENTS_MAINTENANCE', tsde.table_name, tsde.input, tsde.output, null
                 FROM INTERNAL.TASK_SIMPLE_DATA_EVENTS tsde;
-            call internal.set_config(config_key, 'true');
+            call internal.set_config(:config_key, 'true');
+        END IF;
+        commit;
+    end if;
+
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_WAREHOUSE_LOAD_EVENTS')) then
+        BEGIN TRANSACTION;
+        let config_key := (select 'MIGRATION_TASK_WAREHOUSE_LOAD_EVENTS');
+        let config_value text;
+        call internal.get_config(:config_key) into :config_value;
+        if (config_value is null) then
+            -- Different from the previous, carrying table_name into task_log as object_name
+            INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
+                SELECT twle.run, twle.success, 'WAREHOUSE_LOAD_MAINTENANCE', twle.warehouse_name, twle.input, twle.output, null
+                FROM INTERNAL.TASK_WAREHOUSE_LOAD_EVENTS twle;
+            call internal.set_config(:config_key, 'true');
         END IF;
         commit;
     end if;
--- a/bootstrap/012_task_log.sql
+++ b/bootstrap/012_task_log.sql
@@ -6,8 +6,6 @@ $$
      SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
 $$;
 
-let task_log_exists boolean := (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_LOG'));
-
 -- Generic table that tasks should record their execution into.
 CREATE TABLE INTERNAL.TASK_LOG IF NOT EXISTS (task_start timestamp_ltz, success boolean, task_name varchar, object_name varchar,
     input variant, output variant, task_finish timestamp_ltz, task_run_id text, query_id text);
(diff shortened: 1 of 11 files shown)


bootstrap/090_post_setup.sql
```diff
@@ -70,21 +70,84 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_EVENTS_MAINTENANCE
...
 CREATE OR REPLACE TASK TASKS.SIMPLE_DATA_EVENTS_MAINTENANCE
     SCHEDULE = '60 minute'
     ALLOW_OVERLAPPING_EXECUTION = FALSE
     USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "XSMALL"
     AS
-    CALL INTERNAL.refresh_all_simple_tables(true);
+DECLARE
+    task_name text default 'SIMPLE_DATA_EVENTS_MAINTENANCE';
+    task_run_id text default (select INTERNAL.TASK_RUN_ID());
+    query_id text default (select query_id from table(information_schema.task_history(TASK_NAME => 'SIMPLE_DATA_EVENTS_MAINTENANCE')) WHERE GRAPH_RUN_GROUP_ID = :task_run_id  AND DATABASE_NAME = current_database() limit 1);
```
joshelser, line 98: I'd love this to be a UDF; however, because this uses a context_function, it has to live in a task.

We could make a task DAG that ensures a task which creates a common UDF is run before any other task, but this felt overkill for the size of this PR already.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,54 @@
...
+            FROM INTERNAL.TASK_QUERY_HISTORY tqh;
+    end if;
+
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_WAREHOUSE_EVENTS')) then
+        INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
+            SELECT twe.run, twe.success, 'WAREHOUSE_EVENTS_MAINTENANCE', 'WAREHOUSE_EVENTS_HISTORY', twe.input, twe.output, null
+            FROM INTERNAL.TASK_WAREHOUSE_EVENTS twe;
+    end if;
+
+    if (exists (select 1 from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_SIMPLE_DATA_EVENTS')) then
```
rymurr, line 33: not we use `1` here and `*` everywhere else

bootstrap/090_post_setup.sql
rymurr: I have a slight preference for this (and the other similar calls) to live in the `BEGIN`/`END` block. My general rule of thumb was to put only very simple statements into the `DECLAR` block. You do similar on eg line 142

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,54 @@
...
+CREATE TABLE INTERNAL.TASK_LOG IF NOT EXISTS (task_start timestamp_ltz, success boolean, task_name varchar, object_name varchar,
+    input variant, output variant, task_finish timestamp_ltz, task_run_id text, query_id text);
+
+create or replace procedure internal.migrate_task_logs()
+    returns text
+    language sql
+AS
+BEGIN
+    -- If the old task log tables exist, copy the rows into the new task_log table
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_QUERY_HISTORY')) then
```
rymurr, line 21: It looks like we will copy data from the old to the new if the old exists. Where do we delete the old and/or prevent multiple runnings of this creating dub data

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,54 @@
+
+create or replace function internal.task_run_id()
+returns text
+AS
+$$
+     SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
+$$;
+
+let task_log_exists boolean := (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_LOG'));
```
rymurr, line 9: nit: creating the varialbe here an using it so much later in the script was slightly jarring to me

bootstrap/012_task_log.sql
joshelser: Oh, duh. I was thinking about the "next" upgrade and didn't think about the "next-next" upgrade.

joshelser: will truncate old log tables after insert.

bootstrap/012_task_log.sql
joshelser: since truncating the old tables after inserting the rows into the new task_log table, this boolean is just an optimization and not necessary for correctness. Dropped it.

bootstrap/090_post_setup.sql
joshelser: fixed (and swapped the string literal for the `:task_name` variable -- wasn't sure if that was going to work inside the UDTF argument list)

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+create or replace function internal.task_run_id()
+returns text
+AS
+$$
+     SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
+$$;
+
+-- Generic table that tasks should record their execution into.
+CREATE TABLE INTERNAL.TASK_LOG IF NOT EXISTS (task_start timestamp_ltz, success boolean, task_name varchar, object_name varchar,
+    input variant, output variant, task_finish timestamp_ltz, task_run_id text, query_id text);
```
joshelser, line 11: I hadn't included the min/max range of data materialized (e.g. we have query history data from 2023-01-01 00:00:00 to 2023-04-30 05:30:00). because I thought this would be easy to dynamically compute after the fact.

However, it was not as simple as I expected to dynamically construct and run the `select min(..), max(...) from ..` queries. I'm leaning towards to including two new columns (range_min, range_max) into the task_log table.

rymurr: I thoght the min and max were included in the output variant?

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+
+create or replace procedure internal.migrate_task_logs()
+    returns text
+    language sql
+AS
+BEGIN
+    -- If the old task log tables exist, copy the rows into the new task_log table
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_QUERY_HISTORY')) then
+        INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
```
rymurr, line 20: can you put these ( and the below) into a transaction? The insert and teh trunat. Just to ensure they happen atomically

joshelser:
I can't find your comment in Github (hidden now because I pushed a change that affected the line, I think)

> can you put these ( and the below) into a transaction? The insert and teh trunat. Just to ensure they happen atomically

I ended up switching the `truncate table ..` to a `drop table` (we don't need to keep these empty tables laying around). I think your ask is then moot as the DDL is going to auto-commit. LMK if I missed something.

bootstrap/012_task_log.sql
joshelser: Ugh, now Github shows me your comment.

I ended up switching the truncate table .. to a drop table (we don't need to keep these empty tables laying around). I think your ask is then moot as the DDL is going to auto-commit. LMK if I missed something.

bootstrap/012_task_log.sql
joshelser: updated `output` so it includes `range_min` and `range_max`

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
+
+create or replace function internal.task_run_id()
+returns text
+AS
+$$
+     SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
+$$;
+
+-- Generic table that tasks should record their execution into.
+CREATE TABLE INTERNAL.TASK_LOG IF NOT EXISTS (task_start timestamp_ltz, success boolean, task_name varchar, object_name varchar,
```
jacques-n, line 10: should we have oldest and newest data fields in this? I'd rather use structure anywhere we can.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+    returns text
+    language sql
+AS
+BEGIN
+    -- If the old task log tables exist, copy the rows into the new task_log table
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_QUERY_HISTORY')) then
+        INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
+            SELECT tqh.run, tqh.success, 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', tqh.input, tqh.output, null
+            FROM INTERNAL.TASK_QUERY_HISTORY tqh;
+        DROP TABLE INTERNAL.TASK_QUERY_HISTORY;
```
jacques-n, line 23: I don't think we should do this. In this version. I think we should do in future version. If we have a problem and roll back, I think the app would think that a person didn't do full materialization. This would at minimum cause a bunch of extra compute and more likely also cause wrong result (i think). This applies to all the drops/truncates.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+    end if;
+
+    if (exists (select * from information_schema.tables where table_schema = 'INTERNAL' and table_name = 'TASK_WAREHOUSE_EVENTS')) then
+        INSERT INTO INTERNAL.TASK_LOG (task_start, success, task_name, object_name, input, output, task_finish)
```
jacques-n, line 27: Does this code work on a system that doesn't have this table? I'm thinking that you aren't using dynamic sql and I'm not confident on behavior of parsing/if statement behavior in this case.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+            FROM INTERNAL.TASK_SIMPLE_DATA_EVENTS tsde;
+        DROP TABLE INTERNAL.TASK_SIMPLE_DATA_EVENTS;
+    end if;
+END;
+
+-- Try to migrate any previous task log records from the old tables
+call internal.migrate_task_logs();
+
+-- Create specific-purpose views that we had before.
+CREATE OR REPLACE VIEW REPORTING.SIMPLE_DATA_EVENTS_TASK_HISTORY AS SELECT * exclude (task_name) rename (object_name as table_name)
```
jacques-n, line 45: Was this a view before? I'm not convinced that the name is good.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,53 @@
...
+    FROM INTERNAL.TASK_LOG WHERE task_name = 'SIMPLE_DATA_EVENTS_MAINTENANCE';
+CREATE OR REPLACE VIEW REPORTING.WAREHOUSE_LOAD_EVENTS_TASK_HISTORY AS SELECT * exclude (task_name) rename (object_name as warehouse_name)
```
jacques-n, line 47: would be good to also structure input/output variants for each of the types if we're creating these domain-specific views.

bootstrap/018_warehouse_updates.sql
```diff
@@ -34,7 +33,7 @@ begin
     return object_construct('migrate1', migrate1, 'migrate2', migrate2);
 end;
 
-CREATE OR REPLACE PROCEDURE internal.refresh_warehouse_events(migrate boolean) RETURNS STRING LANGUAGE SQL
+CREATE OR REPLACE PROCEDURE internal.refresh_warehouse_events(migrate boolean, input variant) RETURNS OBJECT LANGUAGE SQL
```
jacques-n, line 36: given that variant is input, we should declare (via comment or similar) what the expected fields must be/could be.

bootstrap/018_warehouse_updates.sql
```diff
@@ -82,9 +80,22 @@ BEGIN
...
+                union all
+                select min(session_end) as session_end from INTERNAL_REPORTING_MV.CLUSTER_AND_WAREHOUSE_SESSIONS_COMPLETE_AND_DAILY_INCOMPLETE)
+            );
+            output := OBJECT_CONSTRUCT('oldest_running', :oldest_running, 'newest_completed', :newest_completed, 'attempted_migrate', :migrate, 'migrate', :migrate1, 'migrate_INCOMPLETE', :migrate2,
+                'new_records', :new_records, 'new_INCOMPLETE', :new_INCOMPLETE, 'new_closed', coalesce(:new_closed, 0), 'range_min', :range_min, 'range_max', :newest_completed);
         ELSE
-            insert into INTERNAL.TASK_WAREHOUSE_EVENTS SELECT :dt, true, :input, OBJECT_CONSTRUCT('oldest_running', :oldest_running, 'newest_completed', :newest_completed, 'attempted_migrate', :migrate, 'migrate', :migrate1, 'migrate_INCOMPLETE', :migrate2, 'new_records', 0, 'new_INCOMPLETE', 0, 'new_closed', 0)::VARIANT;
+            let range_min timestamp := (select min(session_end) from (
+                select min(session_end) as session_end from INTERNAL_REPORTING_MV.CLUSTER_AND_WAREHOUSE_SESSIONS_COMPLETE_AND_DAILY
```
jacques-n, line 93: why hand constructing the union as opposed to just using existing object?

bootstrap/019_query_history_updates.sql
```diff
@@ -67,20 +63,32 @@ BEGIN
             let where_clause_complete varchar := (select 'END_TIME <> to_timestamp_ltz(\'' || :newest_completed || '\')');
             let new_closed number;
             call internal.generate_insert_statement('INTERNAL_REPORTING_MV', 'QUERY_HISTORY_COMPLETE_AND_DAILY', 'INTERNAL', 'RAW_QH_EVT', :where_clause_complete) into :new_closed;
-            insert into INTERNAL.TASK_QUERY_HISTORY SELECT :run_id, true, :input, OBJECT_CONSTRUCT('oldest_running', :oldest_running, 'newest_completed', :newest_completed, 'attempted_migrate', :migrate, 'migrate', :migrate1, 'migrate_INCOMPLETE', :migrate2, 'new_records', :new_records, 'new_INCOMPLETE', :new_INCOMPLETE, 'new_closed', coalesce(:new_closed, 0))::VARIANT;
+            -- Figure out the oldest row in the table
+            let range_min timestamp := (select min(end_time) from (
+                select min(end_time) as end_time from internal_reporting_mv.QUERY_HISTORY_COMPLETE_AND_DAILY
```
jacques-n, line 68: as above, why hand construction union?

bootstrap/019_query_history_updates.sql
```diff
@@ -67,20 +63,32 @@ BEGIN
...
+                union all
+                select min(end_time) as end_time from internal_reporting_mv.QUERY_HISTORY_COMPLETE_AND_DAILY_INCOMPLETE
+            ));
+            output := OBJECT_CONSTRUCT('oldest_running', :oldest_running, 'newest_completed', :newest_completed, 'attempted_migrate', :migrate, 'migrate', :migrate1, 'migrate_INCOMPLETE', :migrate2,
+                'new_records', 0, 'new_INCOMPLETE', 0, 'new_closed', 0, 'range_min', :range_min, 'range_max', :newest_completed);
         END IF;
         DROP TABLE RAW_QH_EVT;
         COMMIT;
 
+        CALL INTERNAL.SET_CONFIG('QUERY_HISTORY_MAINTENANCE', CURRENT_TIMESTAMP()::string);
```
jacques-n, line 86: why move so not always happening? Not sure of intention.

bootstrap/090_post_setup.sql
```diff
@@ -70,21 +70,84 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_EVENTS_MAINTENANCE
...
     SCHEDULE = '60 minute'
     ALLOW_OVERLAPPING_EXECUTION = FALSE
     USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "XSMALL"
     AS
-    CALL INTERNAL.refresh_all_simple_tables(true);
+DECLARE
+    task_name text default 'SIMPLE_DATA_EVENTS_MAINTENANCE';
+    task_run_id text default (select INTERNAL.TASK_RUN_ID());
+BEGIN
+    let query_id text := (select query_id from table(information_schema.task_history(TASK_NAME => :task_name)) WHERE GRAPH_RUN_GROUP_ID = :task_run_id  AND DATABASE_NAME = current_database() limit 1);
```
jacques-n, line 99: i think you can pass in the task run id here. or maybe the task root id. Maybe that would be safer. (Imagine our systems where we have several databases with the same task name.)

bootstrap/090_post_setup.sql
```diff
@@ -70,21 +70,84 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_EVENTS_MAINTENANCE
...
+    FOR rowvar IN cur DO
+        let table_name text := rowvar.table_name;
+        let index_col text := rowvar.index_col;
+        BEGIN
+            let start_time timestamp_ltz := (select current_timestamp());
+            let input object := (select output from INTERNAL.TASK_LOG where success and task_name = :task_name and object_name = :table_name order by task_start desc limit 1);
+            INSERT INTO INTERNAL.TASK_LOG(task_start, task_run_id, query_id, input, task_name, object_name) select :start_time, :task_run_id, :query_id, :input, :task_name, :table_name;
+
+            let output variant;
+            CALL INTERNAL.refresh_simple_table(:table_name, :index_col, true, :input) into :output;
```
jacques-n, line 119: It seems like the task log handling above should actually be inside the refresh procedure as opposed to in this overall thing.

bootstrap/090_post_setup.sql
```diff
@@ -70,21 +70,84 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_EVENTS_MAINTENANCE
...
-    CALL INTERNAL.refresh_queries(true);
+DECLARE
+    start_time timestamp_ltz default (select current_timestamp());
+    task_name text default 'QUERY_HISTORY_MAINTENANCE';
+    object_name text default 'QUERY_HISTORY';
+    task_run_id text default (select INTERNAL.TASK_RUN_ID());
+BEGIN
+    let query_id text := (select query_id from table(information_schema.task_history(TASK_NAME => :task_name)) WHERE GRAPH_RUN_GROUP_ID = :task_run_id  AND DATABASE_NAME = current_database() limit 1);
+    let input variant := (select output from INTERNAL.TASK_LOG where success AND task_name = :task_name AND object_name = :object_name order by task_start desc limit 1);
+    INSERT INTO INTERNAL.TASK_LOG(task_start, task_run_id, query_id, input, task_name, object_name) select :start_time, :task_run_id, :query_id, :input, :task_name, :object_name;
```
jacques-n, line 143: as I look at all these inserts/updates... maybe we should make two procedures that encapsulate these behaviors?

bootstrap/012_task_log.sql
joshelser: I thought so, but I'll double check.

bootstrap/012_task_log.sql
joshelser: Yeah, these three (adjacent in this file) views existed prior.

bootstrap/018_warehouse_updates.sql
joshelser: > given that variant is input

The previous code was doing cast to `VARIANT` for no apparent reason. I unwound some of that (after I missed updating a `RETURNS`). I'll just swap them all to be OBJECT since that's what is intended.

bootstrap/018_warehouse_updates.sql
joshelser: The higher-level views which union the complete/incomplete tables also include a filter on `session_type = 'CL'` or `session_type = 'WH'`.

I reckon it would be more correct to deduce the exact range of data available for both cluster and warehouse sessions separately. But, that makes the warehouse materialization case different from the rest. Feasible, but more complex.

bootstrap/012_task_log.sql
joshelser: > If we have a problem and roll back, I think the app would think that a person didn't do full materialization. This would at minimum cause a bunch of extra compute and more likely also cause wrong result (i think).

Yeah, you're exactly right -- it would try to re-materialize things. I'll come up with a different solution that can preserve the old tables.

bootstrap/019_query_history_updates.sql
joshelser: I think how we had it before was confusing.

Being outside of the block looked (at first glance) like we would have always called `SET_CONFIG`. However, we were always `RAISE`'ing in the `EXCEPTION` block, so we would have only called `SET_CONFIG` when we didn't catch an exception.

Putting it inside the `BEGIN` block made it more clear to me that we only call this when the block successfully executes.

bootstrap/018_warehouse_updates.sql
jacques-n: I'm speaking to the fields in the object/variant not the type itself. 

It used to be that the object was entirely inside the procedure. You've moved it to the signature which is why I'm suggesting we document the expected structure.

bootstrap/018_warehouse_updates.sql
jacques-n: Remember that we should be having a row in the user-facing table for each view that we maintain. The user-facing data should be focused on internal tables/views. So in the case of clsuter session and warehouse session, there should be at least two rows (i think more since there is daily and hourly versions I believe)

bootstrap/018_warehouse_updates.sql
joshelser: Ah, I see. Makes sense

bootstrap/018_warehouse_updates.sql
joshelser: > there should be at least two rows (i think more since there is daily and hourly versions I believe)

Yeah, I thought I could get away with a single task_log row for one execution of the warehouse_events_maintenance task, but the distinction between cluster_sessions and warehouse_sessions means I have to change this logic.

The daily and hourly variants are easily included when we're actually creating the materialization status view since they are (usually) backed by the same internal table. Clusters/sessions are the exception. I'll break this out.

bootstrap/012_task_log.sql
joshelser: changed again after Jacques' feedback to not be destructive over the old task log tables.

This sets a value in internal.config to indicate we've migrated the task logs to the new table.

bootstrap/090_post_setup.sql
joshelser: After consolidating the `insert` and `update` into `start_task` and `finish_task`, this feels pretty clean to me and would muddy the waters between the task setup/teardown and task body delineation.

bootstrap/012_task_log.sql
joshelser: With the introduction of `REPORTING.TASK_LOG_HISTORY` (user-facing view for the internal.task_log table), we could just drop these old user-facing views. They are of no value for users of the native app -- just for developers.

bootstrap/012_task_log.sql
joshelser: Resolving as I dropped views. If we want to reintroduce any of these domain-specific views, I think it would be good to be intentional on the schema.

joshelser:
I think I've addressed all of your feedback, @jacques-n. Please take another look when you have the chance.

I need to fix up the unit test failure and add some more while I'm in there.

bootstrap/012_task_log.sql
```diff
@@ -0,0 +1,119 @@
...
+
+create or replace function internal.task_run_id()
+returns text
+AS
+$$
+     SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
+$$;
+
+-- Inserts a row into TASK_LOG with the time the task started, the name of the object being materialized, the graph run ID and query_id for the task.
+create or replace procedure internal.start_task(task_name text, object_name text, start_time timestamp_ltz, task_run_id text, query_id text)
```
joshelser, line 17: just noticed that passing a `timestamp_ltz` through a procedure seems to is shifting by the account's default timezone. passing as text doesn't do this.

jacques-n: let's use timestamp_ntz everywhere.

joshelser: ok

rymurr (approved):
LGTM. I am happy if Jacques is

jacques-n (commented):
Don't wait for me. If Ryan is happy and you addressed changes let's move forward.

joshelser:
Thanks Jacques. No expiration on chance to give more feedback 😆

PR #647, opened 2024-05-01
Selectively reintroduce run_as_app

joshelser:


bootstrap/090_post_setup.sql
```diff
@@ -379,6 +379,28 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_SCHEDULING
     as
     call INTERNAL.UPDATE_WAREHOUSE_SCHEDULES(NULL, NULL);
 
+CREATE OR REPLACE TASK TASKS.SUNDECK_DEBUG
```
jacques-n, line 382: Are we doing this as a task because that's the only way to retrieve the current organization?

I think I'm inclined to add to an existing task as opposed to create a new task.

bootstrap/090_post_setup.sql
```diff
@@ -379,6 +379,28 @@ CREATE OR REPLACE TASK TASKS.WAREHOUSE_SCHEDULING
...
+    ALLOW_OVERLAPPING_EXECUTION = FALSE
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = "X-Small"
+    AS
+BEGIN
+    let org_name text := (select current_organization_name());
+    IF (org_name <> 'SUNDECK') THEN
+        RETURN false;
+    END IF;
+
+    CREATE OR REPLACE PROCEDURE admin.run_as_app(sql string)
```
jacques-n, line 392: With this get the right perms?

bootstrap/090_post_setup.sql
jacques-n, line 382: Let's add a comment to why this exists so that others looking at the code get clarity that this isn't some weird backdoor.

joshelser: > Are we doing this as a task because that's the only way to retrieve the current organization?

Yeah, exactly.

> I think I'm inclined to add to an existing task as opposed to create a new task.

Sure, can do that, instead.

bootstrap/090_post_setup.sql
joshelser: hrm, probably need to add a grant after making it. I bet I have my local app in "DEBUG" mode which is why I didn't get an error. I'll re-test that and add a grant.

joshelser:
> I bet we can read current_organization_name() in setup script

I'll check. My gut reaction is that we can't use context functions in the setup script, but I'm not positive.

> We could then look at the sharing organization of the app we're installing and if the current organization is the same as the installing organization

Yep, I see what you mean. I'll see if that's feasible in the setup script.

bootstrap/090_post_setup.sql
joshelser: I can make the procedure in the task, but the task user (SYSTEM) doesn't have the permission to directly run the grant. 

Maybe the task can call an internal procedure (which is already run as owner) and execute the grant?

moot point if I can make your other suggestion work.

joshelser:
> We could then look at the sharing organization of the app we're installing

I thought we could do something like this:

```
describe application `select current_database()`;
app_org = select "value" from table(result_scan(last_query_id())) where "property" = 'source_organization';
```

But we can't call `describe application` in the setup script. Is there some other way I'm not thinking of to get at the application source information?

jacques-n (dismissed):
bummer we couldn't figure out way to reference the provider_organization_name().

joshelser:
> bummer we couldn't figure out way to reference the provider_organization_name().

Yeah. I'll check with Ryan next time we talk -- maybe there's something I missed which will let us generalize this.

jacques-n (approved):

PR #648, opened 2024-05-02
Create ADMIN.MATERIALIZATION_STATUS view

joshelser:
Remove the task creation (unnecessary now, compared to earlier releases). Update unit tests.

### DIFF
--- a/bootstrap/051_materialization_status.sql
+++ b/bootstrap/051_materialization_status.sql
@@ -1,0 +1,102 @@
+
+create or replace function internal.success_to_status(success boolean) returns text
+as
+$$
+    IFF(success IS NULL, NULL, IFF(success, 'SUCCESS', 'FAILURE'))
+$$;
+
+
+create or replace view admin.materialization_status
+    copy grants
+as
+with task_tables as (
+    select user_schema, user_view, task_name, object_name, task_period_mins from ( values
+        ('REPORTING', 'DBT_HISTORY', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'ENRICHED_QUERY_HISTORY', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'ENRICHED_QUERY_HISTORY_DAILY', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'ENRICHED_QUERY_HISTORY_HOURLY', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'LABELED_QUERY_HISTORY', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'WAREHOUSE_DAILY_UTILIZATION', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+        ('REPORTING', 'WAREHOUSE_HOURLY_UTILIZATION', 'QUERY_HISTORY_MAINTENANCE', 'QUERY_HISTORY', 60),
+
+        ('REPORTING', 'HYBRID_TABLE_USAGE_HISTORY', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'HYBRID_TABLE_USAGE_HISTORY', 60),
+        ('REPORTING', 'LOGIN_HISTORY', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'LOGIN_HISTORY', 60),
+        ('REPORTING', 'MATERIALIZED_VIEW_REFRESH_HISTORY', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'MATERIALIZED_VIEW_REFRESH_HISTORY', 60),
+        ('REPORTING', 'SERVERLESS_TASK_HISTORY', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'SERVERLESS_TASK_HISTORY', 60),
+        ('REPORTING', 'SESSIONS', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'SESSIONS', 60),
+        ('REPORTING', 'TASK_HISTORY', 'SIMPLE_DATA_EVENTS_MAINTENANCE', 'TASK_HISTORY', 60),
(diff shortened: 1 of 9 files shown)


bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,102 @@
+
+create or replace function internal.success_to_status(success boolean) returns text
+as
+$$
+    IFF(success IS NULL, NULL, IFF(success, 'SUCCESS', 'FAILURE'))
+$$;
+
+
+create or replace view admin.materialization_status
```
rymurr, line 9: what is the perf of this? It feels like it would take a while to compile. I assume it will take a long time to execute (with all the joins) if hte table is large

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,102 @@
...
+        ('REPORTING', 'CLUSTER_SESSIONS_DAILY', 'WAREHOUSE_EVENTS_MAINTENANCE', 'CLUSTER_SESSIONS', 60),
+        ('REPORTING', 'CLUSTER_SESSIONS_HOURLY', 'WAREHOUSE_EVENTS_MAINTENANCE', 'CLUSTER_SESSIONS', 60),
+        ('REPORTING', 'WAREHOUSE_SESSIONS', 'WAREHOUSE_EVENTS_MAINTENANCE', 'WAREHOUSE_SESSIONS', 60),
+        ('REPORTING', 'WAREHOUSE_SESSIONS_DAILY', 'WAREHOUSE_EVENTS_MAINTENANCE', 'WAREHOUSE_SESSIONS', 60),
+        ('REPORTING', 'WAREHOUSE_SESSIONS_HOURLY', 'WAREHOUSE_EVENTS_MAINTENANCE', 'WAREHOUSE_SESSIONS', 60)
+    ) as t(user_schema, user_view, task_name, object_name, task_period_mins)
+    UNION ALL
+    -- Expand all pairs of WAREHOUSE_LOAD_MAINTENANCE + warehouse_name (as object_name)
+    select 'REPORTING', 'WAREHOUSE_LOAD_HISTORY', 'WAREHOUSE_LOAD_MAINTENANCE', object_name, 60 from (select distinct object_name from internal.task_log where task_name = 'WAREHOUSE_LOAD_MAINTENANCE')
+), completed_tasks as (
```
rymurr, line 40: I wonder if instead of a bunch of CTEs we can just do new columns. Eg `success is not null as completed, success is nul as running`

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,102 @@
...
+    -- The history rows generated every time a task is run and the table that task is managing
+    select task_start, task_finish, success, input, output, task_name, object_name, IFF(INPUT is null, 'FULL', 'INCREMENTAL') as kind, query_id, range_min, range_max,
+    FROM INTERNAL.TASK_LOG
+    WHERE success is not null
```
rymurr, line 44: what about failures? or are we not differentiating between completed successfully and completed wiht na error?

bootstrap/051_materialization_status.sql
```diff
@@ -0,0 +1,102 @@
...
+), running_tasks as (
+    select task_start, task_finish, success, input, output, task_name, object_name, IFF(INPUT is null, 'FULL', 'INCREMENTAL') as kind, query_id, range_min, range_max,
+    FROM INTERNAL.TASK_LOG
+    WHERE success is null
+), fullmat as (
+    SELECT ct.task_name, ct.object_name, ct.task_start, ct.task_finish, ct.success,
+        output['SQLERRM'] as error_message, ct.range_min, ct.range_max, ct.query_id,
+    FROM completed_tasks ct
+    WHERE (ct.object_name, ct.task_name, ct.task_start) IN (
+        SELECT object_name, task_name, MAX(task_start)
```
rymurr, line 54: this subselect looks odd to me, I don't think I understand the goal well enough to comment on it

bootstrap/051_materialization_status.sql
joshelser: ~900ms for the 20k rows we have in the Sandbox account.

bootstrap/051_materialization_status.sql
joshelser: At this point, we are just looking for rows which are completed (`success is not null`). `fullmat` and `incmat` then pick the most recent (based on task_start) for each of full and incremental materializations

bootstrap/051_materialization_status.sql
joshelser: For each task and full and incremental materialization types, I want to find the row with `max(task_start)` within that task_name and materialization type.

I can add a comment explaining this, but maybe there's a better way?

bootstrap/051_materialization_status.sql
joshelser: I trimmed a bunch of fat off of this as compared to what I started with (specifically, after not using the TASK_HISTORY udtf).

If I can read your mind, this could collapse into something much more simple, like a `select max(task_start) .. group by task_name, object_name, success`?

bootstrap/051_materialization_status.sql
joshelser: Then I reran it later and it was more like 1.8s, roughly split equally between compilation, queuing, and execution.

jacques-n:
I'm doing a rewrite.

deploy/devdeploy.py
```diff
@@ -92,15 +92,17 @@ def _finish_local_setup(cur, database: str, schema: str):
...
+        # SNOWFLAKE_WAREHOUSE_MAINTENANCE is set when TASK_SIMPLE_DATA_EVENTS_MAINTENANCE completes.
         cur.execute(
-            "SELECT * FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE', 'SNOWFLAKE_USER_MAINTENANCE') and value is not null;"
+            """SELECT * FROM internal.config where key in ('WAREHOUSE_EVENTS_MAINTENANCE', 'QUERY_HISTORY_MAINTENANCE',
+            'SNOWFLAKE_USER_MAINTENANCE', 'SNOWFLAKE_WAREHOUSE_MAINTENANCE') and value is not null;"""
         )
 
         rows = cur.fetchall()
 
         # if we have two rows, means materialization is complete
```
jacques-n, line 104: comment looks wrong

jacques-n (approved):

deploy/devdeploy.py
joshelser: fixed

joshelser:
Merging since it was only a test-fix and a comment change since last approval.

PR #651, opened 2024-05-03
Call start_task/finish_task in the task, not in the procedure.

joshelser:
We cannot call task_history from within a stored procedure. Forgot to update this one case because it is in the 092 file rather than 090 like the rest of the instances.

Other things:
- [ ] we can drop `admin.finalize_setup_from_service_account`. will do that in a separate PR
- [ ] we should add a test which verifies that calling `call admin.create_upgrade_check_task();` and `execute task tasks.upgrade_check;` results in the same outcome as if we had called `admin.finalize_setup()` directly (e.g. entries show up in catalog.config). I don't want to run 2 materializations on every PR, so I'm going to see how we add that separately.


vqmarkman (approved):
It looks good to me. I have to also admit that I have a selfish interest on merging this fix today because I want to make sure that rest of the native app tests are running successfully.

joshelser:
Going to merge this to fix some tests that broke after #644. Happy to address feedback you have, Ryan.

PR #654, opened 2024-05-07
Create a task to mark old, incomplete task_log rows as FAILED

joshelser:
* ~~Task runs every hour, looking for rows in internal.task_log where `SUCCESS IS NULL`~~
* ~~If the query_id is still running (blocked, queued, running) or we do not find the row in QUERY_HISTORY, we do nothing to the row.~~
* ~~If we have no query_id, mark the row as failed.~~
* When a task runs, we find all rows for the same task which are `success is null`
* If we have a `query_id` for that row, look up that query in the Snowflake query_history view and set it as the `output`
* Always include a `FORCE_CLOSED: true` attribute in the `output` object.

Sorry for some more procedural code from me. I tried a couple of different ways to run the UPDATE statement with a left-join against QUERY_HISTORY, but they were all brutally slow (cancelled the query after a minute of execution).

This will correct the admin.materialization_status view reporting the `NEXT_` columns as actively running when they are not actually running (easily noticed when the last incremental materialization has a newer start/finish than the `NEXT_` times).

(edited for description accuracy since fa4a45b)

### DIFF
--- a/bootstrap/012_task_log.sql
+++ b/bootstrap/012_task_log.sql
@@ -13,19 +13,26 @@ $$
      SYSTEM$TASK_RUNTIME_INFO('CURRENT_TASK_GRAPH_RUN_GROUP_ID')
 $$;
 
+-- prevent confusion with the overloaded procedure defined below
+drop procedure if exists internal.start_task(text, text, timestamp_ntz, text, text);
+
 -- Inserts a row into TASK_LOG with the time the task started, the name of the object being materialized, the graph run ID and query_id for the task.
-create or replace procedure internal.start_task(task_name text, object_name text, task_run_id text, query_id text)
+create or replace procedure internal.start_task(task_name text, object_name text, task_run_id text, query_id text,
+        task_log_table text default 'INTERNAL.TASK_LOG', query_history_table text default 'SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY')
     returns object
     language sql
 AS
 BEGIN
-    let start_time timestamp_ltz := (select current_timestamp());
-    let input object := (select output from INTERNAL.TASK_LOG where success AND task_name = :task_name AND object_name = :object_name order by task_start desc limit 1);
+    -- Clean up any outstanding rows for this task that have `success is null`. We are relying on the fact that this procedure is called
+    -- when a task begins and all tasks in the NativeApp disallow concurrent execution (else, we may try to update a row for a task still running).
+    CALL INTERNAL.CLOSE_LOG_FOR_TASK(:task_log_table, :query_history_table, :task_name, :object_name);
+
+    let input object := (select output from IDENTIFIER(:task_log_table) where success AND task_name = :task_name AND object_name = :object_name order by task_start desc limit 1);
     -- Set the previous range_min/max when we start a task to avoid a look-back in admin.materialization_status
     let range_min timestamp_ltz := (select :input['range_min']);
     let range_max timestamp_ltz := (select :input['range_max']);
-    INSERT INTO INTERNAL.TASK_LOG(task_start, task_run_id, query_id, input, task_name, object_name, range_min, range_max)
-        select :start_time, :task_run_id, :query_id, :input, :task_name, :object_name, :range_min, :range_max;
+    INSERT INTO IDENTIFIER(:task_log_table)(task_start, task_run_id, query_id, input, task_name, object_name, range_min, range_max)
+				select current_timestamp(), :task_run_id, :query_id, :input, :task_name, :object_name, :range_min, :range_max;
     return input;
 END;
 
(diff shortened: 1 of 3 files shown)


joshelser:
> Could we just add a stored proc to the top of every one of our tasks to mark the previous one as failed if its incomplete? Rather than introducing a new task

That was my original idea, but Jacques had specifically suggested (offline) that we do it as a separate task, checking against QUERY_HISTORY, too.

It's yet another task to manage, but it is nice that we don't complicate the normal logic of each task.

jacques-n:
I believe tasks expose manage to consumers. Is that the case? If so, what sets of properties can be altered with that perm and which cannot?

joshelser:
> Is that the case? If so, what sets of properties can be altered with that perm and which cannot?

That's a great question. Looks like as a normal user, you cannot alter the task.

```
josh@sundeck.io#COMPUTE_WH@OPSCENTER.(no schema)>alter task tasks.upgrade_check set schedule = '60 minutes';
003001 (42501): SQL access control error:
Insufficient privileges to operate on task 'UPGRADE_CHECK'
josh@sundeck.io#COMPUTE_WH@OPSCENTER.(no schema)>alter task tasks.upgrade_check unset ALLOW_OVERLAPPING_EXECUTION;
003001 (42501): SQL access control error:
Insufficient privileges to operate on task 'UPGRADE_CHECK'
```

But the user can `suspend`, `resume`.

rymurr:
Why would we want to give the user the option to suspend or resume a task that is maintaining task log state? If I understand correctly all we want is to mark failed tasks as failed and the very fact that a task is running is indicative of the previous one having finished.

It doesn't make much sense to me to complicate things w/ this extra task

joshelser:
> Why would we want to give the user the option to suspend or resume a task that is maintaining task log state?

I was just trying to understand what things a NativeApp owner can change about a task vs. the things they cannot change.

1. A user can suspend or resume a task if they so choose.
2. All other attributes of a task cannot be changed (that I've tried, so far).

Going one step further, that means a user could not "un-set" the invariant that all of the tasks are created as `ALLOW_OVERLAPPING_EXECUTION = FALSE`. Finally, implying that the solution of doing this work when each task starts will be effective given the scope of what a user can do on their own.

I believe this is what Jacques was hinting at.

jacques-n:
Yes Josh, that was what I was asking.

There are two different design choices from my pov...

1. Who notices an open task: self observed or external observed

2. What is the correction: blindly mark failed vs look up specific failure and record. 

For 1, no strong opinion. 
For 2, strong preference for specific failure. 

Additionally, whenever the cleanup op needs to cleanup we should consider it a bug.

joshelser:
> I am doing one more manual test locally, which is injecting a failure into the body of one of these tasks and making sure the "catch" block of the procedure is correctly closing the record.

Confirmed that a task body failing does appropriately result in a `SUCCESS=FALSE` row that is reported through materialization_status as expected.

bootstrap/012_task_log.sql
```diff
@@ -152,3 +160,47 @@ CREATE OR REPLACE VIEW ADMIN.TASK_LOG_HISTORY AS SELECT * FROM INTERNAL.TASK_LOG
...
+    end for;
+
+    let o object := OBJECT_CONSTRUCT('success', True, 'rows_closed', :rows_closed);
+    SYSTEM$LOG_INFO(:o);
+    return o;
+EXCEPTION
+    WHEN OTHER THEN
+        SYSTEM$LOG_ERROR(OBJECT_CONSTRUCT('error', 'Exception trying to close outstanding task rows.', 'SQLCODE', :sqlcode, 'SQLERRM', :sqlerrm, 'SQLSTATE', :sqlstate));
+        return OBJECT_CONSTRUCT('SQLCODE', :sqlcode, 'SQLERRM', :sqlerrm, 'SQLSTATE', :sqlstate);
+END;
```
rymurr, line 206: The procedural version seems heavy and hard for me to parse. I've got:

```
update <insert update here> from
with foo as (
select current_timestamp as task_start, 'foo' as task_run_id, '01b2438c-0904-46df-0062-5003077b8cfe' as query_id,
'Could not find query history for task=' || 'task_name' || ', object=' || 'objecT_name' || ', task_run_id=' || task_run_id || ', query_id=' || COALESCE(query_id, 'NULL') as msg, OBJECT_CONSTRUCT('FORCE_CLOSED', TRUE) output
), bar as (
select object_construct(q.*) as qh, task_start, task_run_id, msg, output from snowflake.account_usage.query_history q inner join foo on q.query_id = foo.query_id
)
select case when qh is null then objecT_insert(output, 'details', msg) else  dryan.internal.merge_objects(output, qh) end as output from bar;
```

working. I think the changes would be:
* use right outer join
* i think more things need to be passed through the CTE
* finish the update statement

joshelser: Thanks for the snippet. I worry that this is going to be slow as molasses because querying the SecureView is going to prevent the predicate pushdown from happening.

When I run your example (with a constant value for the query_id), I can see that only one row is returned from query_history in a few seconds. However, if I replace `foo` with a query that reads a query_id from a table, then the query profile shows scanning thousands of rows of query_history, running for over a minute without completing.

Two options I'm looking at:
* I can try to clean up the creation of the `output` to be a little less ugly, and we continue to go down the path I started to on this PR
* We stop trying to do the query_history aggregation at task execution time, and create a 2nd view in the native app which does the join against our materialized view query history. Something like
```
create view admin.materialization_failures as 
with failures as (
  select * from admin.task_log_history where success is null output:forced_closed = true
) select *, object_construct(eqh) as query_history_details from failures
left join reporting.enriched_query_history eqh on failures.query_id = eqh.query_id;
```

joshelser: I ended up doing the first option from my last comment as it was most straightforward to implement without writing an entirely new piece to this PR.

I'm not particularly surprised about the secure view performance being bad for the join case (like you originally suggested) and a `from query_history where query_id in (select query_id from tasks_needing_closed)`, but it's frustrating none the less.

The current approach in here runs in ~4seconds which feels reasonable.

rymurr: Is it that the predicate isn't being pushed down or is it that we don't have a good predicate? Is it still slow if we limit by time? eg `from query_history where query_id in (select query_id from tasks_needing_closed) and start_time > current_timestamp - interval '1 day'`

joshelser: > Is it that the predicate isn't being pushed down or is it that we don't have a good predicate?

Well, I can't say that I know for certain, but it feels like the optimizer doing something in one case and not the other (e.g. constant string query_id -> fast and reads one row. query_id taken from CTE expression -> slow, reading many rows).

> Is it still slow if we limit by time

I'll check.

joshelser: > Is it still slow if we limit by time

yeah, looks like filtering on time does greatly help.

bootstrap/012_task_log.sql
```diff
@@ -152,3 +160,52 @@ CREATE OR REPLACE VIEW ADMIN.TASK_LOG_HISTORY AS SELECT * FROM INTERNAL.TASK_LOG
...
+CREATE OR REPLACE PROCEDURE INTERNAL.CLOSE_LOG_FOR_TASK(task_log_table text, query_history_table text, task_name text, object_name text)
+    RETURNS OBJECT
+    LANGUAGE SQL
+    Comment = 'Close any outstanding task rows that have not been updated since the given timestamp. If a query_id is present, the query_history row is set as the output'
+AS
+BEGIN
+    let outstanding_rows resultset := (select task_start, task_run_id, query_id from IDENTIFIER(:task_log_table) where success is null and task_name = :task_name and object_name = :object_name);
+    let cur cursor for outstanding_rows;
+    let rows_closed integer := 0;
+    for r in cur do
```
rymurr, line 173: sorry to be a pain bu tI am still striugglign as to why we need the for loop. I am really trying to get away from that.

joshelser: Not a pain, I didn't push my local changes after yesterday's distraction. Local change makes this whole proc an update stmt. Sorry for the confusion

joshelser: pushed the change

rymurr (approved):

joshelser:
Just fixed a test since last approval. Merging before it becomes in conflict again.

PR #655, opened 2024-05-08
Change WAREHOUSE_LOAD_HISTORY to run every 7 hours.

joshelser:
Every 7 hours instead of 8, I'm assuming, so we still get the 5 second windows instead of 5 minute windows we get when querying the table function for more hours.

Additionally, for incremental materializations, we will only make one table function query for each task execution (where an 8hour schedule may result in us making two queries to the table function).


bootstrap/023_simple_materializations.sql
```diff
@@ -128,9 +128,9 @@ begin
         {
           'warehouse_name': :warehouse_name,
           'start_time': dateadd(hours, value, :start_time)::text,
-          'end_time': dateadd(hours, value+8, :start_time)::text
+          'end_time': dateadd(hours, value+7, :start_time)::text
         })
-    from table(flatten(input=>array_generate_range(0, datediff(hours, :start_time, current_timestamp()), 8))) f);
+    from table(flatten(input=>array_generate_range(0, datediff(hours, :start_time, current_timestamp()), 7))) f);
```
jacques-n, line 133: if the logic worked before, it doesn't seem like this needs to change. we only need to change the schedule, no?

Also, I think you need to update the task log summary view.

joshelser: > if the logic worked before, it doesn't seem like this needs to change. we only need to change the schedule, no?

I was waffling on it, honestly. Happy to walk this back.

> Also, I think you need to update the task log summary view.

Ah, and now I know what you meant! :)

jacques-n (approved):

PR #656, opened 2024-05-08
Change SIMPLE_DATA_EVENTS_MAINTENANCE to run every 12hrs

joshelser:
Also, move the refreshing of internal.sfwarehouses into WAREHOUSE_LOAD_MAINTENANCE (since that task reads those warehouses).


joshelser:
need to fix materialization_status here, too

bootstrap/090_post_setup.sql
```diff
@@ -387,6 +384,9 @@ DECLARE
 BEGIN
     let query_id text := (select query_id from table(information_schema.task_history(TASK_NAME => :task_name, ROOT_TASK_ID => :root_task_id)) WHERE GRAPH_RUN_GROUP_ID = :task_run_id  AND DATABASE_NAME = current_database() limit 1);
 
+    -- Refresh the warehouses prior to refreshing the warehouse load history
+    call internal.refresh_warehouses();
```
jacques-n, line 388: if we're calling here, can we remove any other calls/tasks that do this?

joshelser: > if we're calling here, can we remove any other calls/tasks that do this?

I think it was just the one I removed in this patch, but i'll dbl check.

jacques-n: wait, why don't you move the operation into warehouse_load_history? Isn't that where we need it?

joshelser: Sorry, maybe I'm missing something: that's what I did here. I moved it from SIMPLE_DATA_EVENTS_MAINTENANCE to WAREHOUSE_LOAD_MAINTENANCE. We will refresh our table of warehouses and then we refresh the WLH for each wh in that table.

jacques-n: ah... misleading github snippet. makes sense.

jacques-n (approved):

rymurr (approved):

rymurr:
I am going to merge to get a build out to our internal accounts.

PR #659, opened 2024-05-08
Move task log migration into each task

joshelser:
## Description

After introducing the task_log table, Ryan found that there were duplicated query history records. Upon further investigation, we found that this happened until the native app successfully ran the FINALIZE_SETUP (via the UPGRADE_CHECK task). Until that point, the old task definitions continued to materialize new rows but log their progress to the old tables. Once the new task definitions were created by FINALIZE_SETUP, the system worked as intended (but we were left with duplicated rows).

This PR:
* Ensures that the old task definitions do not continue to run against the old, non-consolidated task log tables.
* Drops an UPDATE statement that was fixing an earlier issue where TASK_FINISH wasn't filled in on migration since we cannot trust the current state of the consolidated task_log
* Creates a full history of task_log rows for the "meta" entries of CLUSTER_SESSIONS and WAREHOUSE_SESSIONS rather than just the full materializations (in case we need to do analysis later).

A reasonable period of time after the new native app release, we can drop the migration logic from the tasks.

## Checklist:


test/unit/test_materialization.py
```diff
@@ -238,71 +242,143 @@ def test_materialization_status_structure(conn):
...
 
         for row in rows:
             assert row["USER_SCHEMA"] is not None
             assert row["USER_VIEW"] is not None
             if row["USER_VIEW"] == "WAREHOUSE_LOAD_HISTORY":
                 assert row["PARTITION"] is not None
-            for c in ["RANGE_START", "RANGE_END"]:
+
+            assert row["LAST_FULL_STATUS"] == "SUCCESS"
+            for c in ["LAST_FULL_START", "LAST_FULL_END"]:
```
joshelser, line 255: Updated these checks to verify this as we have downstream checks that ensure we have a full materialization

test/unit/test_materialization.py
```diff
@@ -238,71 +242,143 @@ def test_materialization_status_structure(conn):
...
                 assert row[c] is not None, f"Expected field {c} to be not null: {row}"
                 assert isinstance(row[c], datetime.datetime)
-            assert (
-                row["RANGE_START"] <= row["RANGE_END"]
-            ), "Range end should never be greater than start"
+            # If we have a range of data, the start should always be less than or equal to the end.
+            if row["RANGE_START"] and row["RANGE_END"]:
+                assert (
+                    row["RANGE_START"] <= row["RANGE_END"]
+                ), "Range end should never be less than range start"
```
joshelser, line 262: We may not have data for a given kind of materialization and that's OK.

rymurr (approved):

PR #664, opened 2024-05-10
Materialize qtag filter column

rymurr:
# Pull Request Template

## Description

Add qtag_filter as a materialized column rather than a view column. This adds to the base table and also adds a backfill mechanism. The backfill mechanism runs inline with the query history maintenance task and updates a day per run starting at the most recent day and moving backwards in time. The slow update smooths out the extra cost of the update and keeps the update size relatively small to prevent unnecessary locking. With a 1 hour frequency it will take 2 days to cover all the time range that our reports usually cover and about 15 days to go back a year.

We incur an extra cost by doing this. It takes 5-20 seconds to update a day on our accounts leading to max 2 hrs of extra compute ($4.5 in serverless tasks). Once the backfill is complete the update takes <1s so is negligible once backfill is done.

We will remove the backfill task after about a month.

## Checklist:


bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,18 @@ BEGIN
         call internal.set_config(:key, 'true');
     end if;
 END;
+
+create or replace procedure internal.update_qtag_day()
+returns number
+language sql
+comment = 'materialize qtag_filter for a single day until fully backfilled'
+as
+begin
```
joshelser, line 179: nit: some whitespace would go a long way 😄

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,18 @@ BEGIN
...
+update internal_reporting_mv.query_history_complete_and_daily set qtag_filter=tools.qtag_to_map(qtag)
+where start_time::date = (select max(start_time::date) from internal_reporting_mv.query_history_complete_and_daily where qtag_filter is null and qtag is not null) and qtag_filter is null and qtag is not null;
+let updates number := (select $1 from table(result_scan(last_query_id())));
+update internal_reporting_mv.query_history_complete_and_daily_incomplete set qtag_filter=tools.qtag_to_map(qtag)
+where start_time::date = (select max(start_time::date) from internal_reporting_mv.query_history_complete_and_daily_incomplete where qtag_filter is null and qtag is not null) and qtag_filter is null and qtag is not null;
+updates := (select $1 from table(result_scan(last_query_id()))) + :updates;
+return updates;
+end;
```
joshelser, line 187: should add an `EXCEPTION` block here in case something goes awry since there isn't one in the caller.

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,18 @@ BEGIN
         call internal.set_config(:key, 'true');
     end if;
 END;
+
+create or replace procedure internal.update_qtag_day()
```
joshelser, line 174: Also, would be great to have a simple unit test to verify this proc, even if it is naive. We likely wouldn't notice breakages here until the app is deployed, given grants, and finalize_setup is run.

bootstrap/017_query_history.sql
rymurr: fixed

bootstrap/017_query_history.sql
rymurr: added

bootstrap/017_query_history.sql
rymurr: added soemthign silly, at least it will check that things work

test/unit/test_materialization.py
```diff
@@ -473,3 +474,21 @@ def test_close_stale_task_log(conn):
...
+"""
+        res = cur.execute(sql.format(tbl_suffix="")).fetchall()
+        assert len(res) == 1
+        updated_rows = res[0].get("number of rows updated")
+        assert updated_rows is not None
+
+        res = cur.execute("CALL internal.update_qtag_day()").fetchall()
+        assert len(res) == 1
+        assert len(res[0]) == 1
+        assert abs(res[0]["UPDATE_QTAG_DAY"] - updated_rows) < updated_rows * 0.2
```
joshelser, line 494: I'm struggling with what this row is asserting -- we're checking that one execution of this procedure is never updating more than 20% of the total number of rows from today?

Doesn't that make this dependent on the number of query history rows that have qtags? Feels brittle.

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,44 @@ BEGIN
...
+
+UPDATE internal_reporting_mv.query_history_complete_and_daily_incomplete
+SET    qtag_filter = tools.qtag_to_map(qtag)
+WHERE  start_time :: DATE = (SELECT Max(start_time :: DATE)
+                             FROM
+              internal_reporting_mv.query_history_complete_and_daily_incomplete
+                             WHERE  qtag_filter IS NULL
+                                    AND qtag IS NOT NULL)
+       AND qtag_filter IS NULL
+       AND qtag IS NOT NULL;
```
joshelser, line 205: nit: had to look twice to make this update statement was the same as the one above.

joshelser (dismissed):

test/unit/test_materialization.py
joshelser: Ok, I get it now -- verifying that the two values are very close together (i.e. we updated all of the rows we expected to update)

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,44 @@ BEGIN
...
+AND    qtag_filter IS NULL
+AND    qtag IS NOT NULL;
+
+let updates number :=
+(
+       SELECT $1
+       FROM   TABLE(result_scan(last_query_id())));
+
+
+UPDATE internal_reporting_mv.query_history_complete_and_daily_incomplete
```
jacques-n, line 197: i don't think we should update incomplete. It gets dropped and recreated every hour so no point.

bootstrap/090_post_setup.sql
```diff
@@ -205,6 +205,11 @@ BEGIN
     let output object;
     CALL INTERNAL.refresh_queries(true, :input) into :output;
 
+    let qtag_updates number;
+    call internal.update_qtag_day() into :qtag_updates;
+
+    output := OBJECT_INSERT(:output, 'updated_query_tag_filters', :qtag_updates);
```
jacques-n, line 211: Change to qtag. Query tags are something else.

bootstrap/090_post_setup.sql
```diff
@@ -205,6 +205,11 @@ BEGIN
     let output object;
     CALL INTERNAL.refresh_queries(true, :input) into :output;
 
+    let qtag_updates number;
+    call internal.update_qtag_day() into :qtag_updates;
```
jacques-n, line 209: let's add comment about plan to remove 15 days after deploy.

rymurr:
@jacques-n and @joshelser i've taken a slightly different approach. Please let me know what you think

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,54 @@
...
+create or replace procedure task_queue.run_tasks()
+returns varchar
+language sql
+as
+begin
+show warehouses;
+show tasks in schema task_queue;
+let res resultset := (
+with tlist as (
+select "id" as id, "name" as name from table(result_scan(last_query_id())) where "name" <> 'run_tasks'
```
joshelser, line 11: Need to pass `database_name` and `schema_name` through, else your query of `account_usage.task_history` is going to potentially overmatch to tasks outside of this database.

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,54 @@
...
+),
+wlist as (
+select count(*) > 0 as wh_exists from table(result_scan(last_query_id(-2))) where "name" = 'SUNDECK_TASK_QUEUE_WH'
+),
+thist as (
+    select tlist.name, state from tlist left outer join account_usage.task_history th on tlist.id = th.task_id
+), tstate as (
+    select name, state, row_number() over (partition by name order by state desc) as rn from thist
```
joshelser, line 19: I'm a bit confused by this partitioning -- `order by state desc` seems odd. Is that what you intended?

It looks like it would be returning the oldest record for each task from the lexicographically-largest state (often, `SUCCEEDED`)

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,54 @@
+
+create or replace procedure task_queue.run_tasks()
+returns varchar
+language sql
+as
+begin
```
joshelser, line 6: Probably want to generate a row in internal.task_log for this "meta" task?

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,59 @@
...
+
+create or replace procedure task_queue.create_tasks()
+returns varchar
+language sql
+comment='Create tasks for the task queue, this has to be done after we have permissions in the native app'
+as
+begin
+create task if not exists task_queue.run_tasks_task
+schedule = '1440 minute'
+allow_overlapping_execution=FALSe
```
joshelser, line 43:
```suggestion
allow_overlapping_execution=FALSE
```

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,59 @@
...
+USER_TASK_TIMEOUT_MS = 30000
+as
+call task_queue.run_tasks();
+
+create task if not exists task_queue.create_warehouse
+as
+call task_queue.create_warehouse();
+end;
+create or replace procedure task_queue.create_warehouse()
```
joshelser, line 52:
```suggestion

create or replace procedure task_queue.create_warehouse()
```

rymurr: fixed

bootstrap/013_task_queue.sql
rymurr: fixed

bootstrap/013_task_queue.sql
rymurr: fixed

bootstrap/013_task_queue.sql
```diff
@@ -14,7 +14,7 @@ wlist as (
 select count(*) > 0 as wh_exists from table(result_scan(last_query_id(-2))) where "name" = 'SUNDECK_TASK_QUEUE_WH'
 ),
 thist as (
-    select tlist.name, state, completed_time from tlist left outer join account_usage.task_history th on tlist.id = th.task_id
+    select tlist.name, state, completed_time from tlist left outer join account_usage.task_history th on tlist.id = th.task_id where th.database_name = current_database() and schema_name='TASK_QUEUE'
```
joshelser, line 17: I think this will fail -- can't use `current_database` in an owners-rights procedure. Pass in the database name from `task_queu.run_tasks_task`

bootstrap/013_task_queue.sql
joshelser: sorry, would want to add in to `task_queue.run_tasks_task`, not here.

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,32 @@ BEGIN
...
+            distinct start_time::date as day
+        FROM
+            internal_reporting_mv.query_history_complete_and_daily
+        WHERE
+            qtag_filter IS NULL
+            AND qtag IS NOT NULL
+        ORDER BY day desc
+    );
+    let cur cursor for res;
+    for r in cur do
```
jacques-n, line 192: I'd put a note here as to why we're doing this a day at a time wrt locking.

bootstrap/017_query_history.sql
```diff
@@ -166,3 +170,32 @@ BEGIN
...
+create or replace procedure internal.update_qtag_day()
+returns varchar
+language sql
+comment = 'materialize qtag_filter for a single day until fully backfilled'
+as
+begin
+
+    let res resultset := (
+        SELECT
+            distinct start_time::date as day
```
jacques-n, line 183: you can use st_day here and below. it's a partition column. it probably isn't logically different but makes thing clearer.

bootstrap/013_task_queue.sql
```diff
@@ -0,0 +1,61 @@
...
+), tstate as (
+    select name, state, row_number() over (partition by name order by completed_time desc) as rn from thist
+), alltasks as (
+select 'task_queue.' || name as name, state from tstate where rn = 1)
+select case when state = 'SUCCEEDED' then 'drop task if exists ' || name || ';'
+when wh_exists then 'begin\nalter task '||name||' set warehouse = SUNDECK_TASK_QUEUE_WH;\nexecute task '|| name ||';\nend;'
+else 'begin\nalter task '||name||' unset warehouse;\nexecute task '|| name ||';\nend;'
+end as cmd from alltasks, wlist where state <> 'RUNNING'
+);
+let cur cursor for res;
```
jacques-n, line 27: can you move to sql and just get back one block?

bootstrap/013_task_queue.sql
rymurr: good shouit

bootstrap/017_query_history.sql
rymurr: done

bootstrap/017_query_history.sql
rymurr: using `st_peroid` which is what you meant I think. Its `datetrunc(day, start_time)`

bootstrap/013_task_queue.sql
rymurr: fixed

bootstrap/013_task_queue.sql
rymurr: done

rymurr:
I;ve moved this to: https://github.com/sundeck-io/native-app/pull/1 with the exact same code/commits.

PR #666, opened 2024-05-13
Grant Create Warehouse Role to Native App

anshuldata:
# Pull Request Template

## Description

* Grant Create Warehouse role to App.
* No extra privilege needs to be given for drop. It is because creator gets an ownership on object

## Checklist:


bootstrap/001_schemas_and_roles.sql
```diff
@@ -35,3 +35,5 @@ GRANT USAGE ON SCHEMA ADMIN TO APPLICATION ROLE ADMIN;
 GRANT USAGE ON SCHEMA ADMIN TO APPLICATION ROLE SUNDECK_SERVICE_ROLE;
 
 create view catalog.config if not exists as select * from internal.config;
+
+GRANT CREATE WAREHOUSE ON ACCOUNT TO APPLICATION ROLE ADMIN;
```
pravindra, line 39: also, ALTER/DROP ?

pravindra (approved):

bootstrap/001_schemas_and_roles.sql
anshuldata: I think we don't need it 

https://docs.snowflake.com/en/user-guide/security-access-control-privileges
`OWNERSHIP is a special privilege on an object that is automatically granted to the role that created the object`

anshuldata: I verfied in Sf, with only Create Warehouse privliege, I could delete, created warehouse

anshuldata:
App deploy is failing with 
`File: 001_schemas_and_roles.sql
[Uncaught exception of type 'STATEMENT_ERROR' on line 39 at position 0 : Grant not executed: Insufficient privileges.] was primary error. `

I think we have to ask permission from user similar to manage warehouse ([here](https://github.com/sundeck-io/OpsCenter/blob/76abd5b4c1136b7c36e3ecdd1c126f3907d458f5/app/ui/setup.py#L25)). Is it correct understanding  @rymurr ?

pravindra:
> I think we have to ask permission from user similar to manage warehouse ([here](https://github.com/sundeck-io/OpsCenter/blob/76abd5b4c1136b7c36e3ecdd1c126f3907d458f5/app/ui/setup.py#L25)). Is it correct understanding @rymurr ?

yes.

bootstrap/001_schemas_and_roles.sql
pravindra: ok - also try changing the tag of warehouse, adding it to a resource monitor. see [here](https://docs.snowflake.com/en/user-guide/security-access-control-privileges#virtual-warehouse-privileges)

rymurr: did you try this in a full deployment? I don't think this will work (or is required) if the native app has the correct permissions

anshuldata: yes, this won't be needed.
I realized that this won't work with failed PR test. I am fixing it to ask permission from user as discussed on slack

anshuldata: fixed warehouse permission

anshuldata: For tag -- We can't use tag in standard edition
Resource monitor -- only account admin can create it

rymurr (approved):

